PR #22, opened 2017-06-09
Zip and zipWith

DzyubSpirit:


lib/functional.js
```diff
@@ -4,4 +4,30 @@ module.exports = (api) => {
 
   api.common.curry = (fn, ...args) => fn.bind(null, ...args);
 
+  api.common.zip = (...arrs) => {
```
tshemsedinov: Please add short description in sources and we need tests which can be used as examples

lib/functional.js
```diff
@@ -4,4 +4,30 @@ module.exports = (api) => {
...
+    if (arrs.length === 0) return [];
```
tshemsedinov: `arrs` or `args` ?

tshemsedinov: if arguments will be only arrays I'd prefer name `arrays`

lib/functional.js
```diff
@@ -4,4 +4,30 @@ module.exports = (api) => {
...
+
+    let minLen = arrs[0].length;
+    let i;
+    for (i = 1; i < arrs.length; i++) {
+      minLen = Math.min(arrs[i].length, minLen);
```
tshemsedinov: map/reduce implementation will be so short and beautiful here but need to test performance

lib/functional.js
```diff
@@ -4,4 +4,30 @@ module.exports = (api) => {
...
+      res[i] = new Array(arrs.length);
+      let j;
+      for (j = 0; j < res[i].length; j++) {
+        res[i][j] = arrs[j][i];
+      }
+    }
+    return res;
+  };
+
+  api.common.zipWith = (fn, ...arrs) => (
```
tshemsedinov: Maybe we can use compose here

DzyubSpirit: As I see it can be rewrited as:
```javascript
const { compose, zip, curry } = api.common;
api.common.zipWith = (fn, ...arrays) => compose(zip, curry(args => fn(...args)))(...arrays);
```
Do you know how to short that because I really like to use `compose` instead chaining

DzyubSpirit:
@tshemsedinov Added tests and documentation

lib/functional.js
tshemsedinov: I can't find better solution to compose chain then use lambda:
```
const composedFunction = data => data.filter(f1).map(f1);
```

tshemsedinov: But it is interesting question, let's think...

tshemsedinov: See #37

tshemsedinov:
You can squash and land and chain composition will be next step @DzyubSpirit

DzyubSpirit:
Landed in fbecae4d5d1189b, 4c03b9732af6f4, 69c959c38ca9fd70e715

Issue #16, opened 2017-06-06
Implement curryTwice


DzyubSpirit:
`curryTwice` transforms function in so way that it then could be execute twice.
```javascript
const curryTwice = api.common.curryTwice;
const asAsync = api.metasync.asAsync;
const sum = (x, y) => (x + y);
const result = curryTwice(sum)(4)(5);
// result: 9
const dur1 = asAsync(fs.readFile, 'config1.json').fmap(JSON.parse).fmap(x => x.duration);
const dur2 = asAsync(fs.readFile, 'config2.json').fmap(JSON.parse).fmap(x => x.duration);
const result2 = api.metasync.asAsync(curryTwice(sum)).ap(dur1).ap(dur2);
result2(console.log);
// prints sum of two durations
```

PR #24, opened 2017-06-09
Implement curryTwice

DzyubSpirit:
Closes: #16

### DIFF
--- a/lib/functional.js
+++ b/lib/functional.js
@@ -4,4 +4,6 @@ module.exports = (api) => {
 
   api.common.curry = (fn, ...args) => fn.bind(null, ...args);
 
+  api.common.curryTwice = fn => api.common.curry(api.common.curry, fn);
+
 };


DzyubSpirit:
A case pushing me to implement that functions is using common function to combine two asynchronous results like that:
```javascript
const af1 = api.common.curry(fs.readFile, 'file1.txt');
const af2 = api.common.curry(fs.readFile, 'file2.txt');
const combine = (file1, file2) => (...);
af2.ap(af1.fmap(twiceCurry(combine)))((err, res) => (
...
));
```
But `fullCurry` got this case. `fullCurry` doesn't cover all cases of `twiceCurry` but I don't know real case for implement it after `fullyCurry` proposal.

tshemsedinov:
@DzyubSpirit need more cases, because above can be easily solved with asynchronous composition.

DzyubSpirit:
Landed in 2fea99c

Issue #31, opened 2017-06-12
Implement Either type


DzyubSpirit:
[safeRequire](https://github.com/metarhia/jstp/pull/216) function in metarhia/jstp can be implemented using Either type which will be used commonly in projects I think.

DzyubSpirit:
@nechaido Could you describe please why you prefer [array](https://github.com/metarhia/jstp/pull/216#issuecomment-307629957)?

nechaido:
It is more consistent this way, IMHO. And also seems like overkill in that case.

DzyubSpirit:
@nechaido Could you please give an example of implementation with tuples for that case for comparision?

nechaido:
@DzyubSpirit here it is:
```js
const getTuple = () => [null, 5];

const [error, data] = getTuple();

if (error) {
  console.log('Error');
} else {
  console.log(data);
}
```

DzyubSpirit:
@nechaido 
```javascript
const getEither = () => ({ isLeft: false, val: 5 }); // can be eitherRight(5), but it's not fair because tupleData can be implemented

const { isLeft, val } = getEither();

if (isLeft) {
 console.log('Error');
} else {
  console.log(val);
}
```
Seems pretty same, but what I really ask to implement is that [case](https://github.com/metarhia/jstp/blob/master/lib/record-serialization.js#L15)

DzyubSpirit:
### Comparing object/hash and tuple/array implementations
#### Conclusion
Tuples are compatible with errback contract for callbacks unlike hash. Hash is better checking for left type value as primitive type (but isn't common use case). Other differences aren't important in my opinion.
Therefore I propose tuples. 
#### Initialization
```javascript 
const eitherValTuple = [err, val]; // for tuple
const eitherValObject = { isLeft, val }; // for object
const eitherValBoth = makeEither(err, val); // for both
```
Syntax for initializing tuples is better because it don't need `err` and `val` names against object needing `isLeft` and `val` variable names or worse syntax `{ isLeft: false, val:  'data' }`.
#### Checking whether either left or right type
```javascript
const [err, val] = eitherValTuple;
if (err !== null && err !== undefined) {
  // if it's left type
} else {
  // if it's right type
}
const { isLeft, val } = eitherValObject;
if (isLeft) {
  // if it's left type
} else {
  // if it's right type
}
```
Syntax for object is better.
#### Implementing methods
```javascript
// for tuples
const makeEither = (...) => {
  ...
[...]

PR #33, opened 2017-06-13
Implement Either type as object

DzyubSpirit:
Refs: #31

### DIFF
--- a/lib/functional.js
+++ b/lib/functional.js
@@ -32,9 +32,6 @@ module.exports = (api) => {
     value !== undefined && value !== null ? fn(value) : defVal
   );
 
-  // Either can be used for saving either left type or right type value.
-  // Left type is commonly used for additional information like error.
-  //
   function Either(
     // Either can be used for saving either left type or right type value.
     // Left type is commonly used for additional information like error.
@@ -43,7 +40,7 @@ module.exports = (api) => {
     // Returns tuple containing left type value or null as first element
     // and right type value or undefined as second element
   ) {
-    if (left) {
+    if (left !== null && left !== undefined) {
       this.isLeft = true;
       this.val = left;
     } else {
@@ -72,18 +69,17 @@ module.exports = (api) => {
     ...args // arguments for concatenation
     // Returns result of concatenation of args
   ) => {
-    switch (args.length) {
-      case 0: throw new Error('altConcat takes one or more arguments');
-      case 1: return args[0];
-      default: {
-        let res = args[0];
-        let i;
-        for (i = 1; i < args.length; i++) {
-          res = api.common.altAppend(res, args[i]);
-        }
-        return res;
-      }
+    if (args.length === 0) {
+      throw new Error('altConcat takes one or more arguments');
     }
+    if (args.length === 1) return args[0];
+
+    let res = args[0];
+    let i;
+    for (i = 1; i < args.length; i++) {
+      res = api.common.altAppend(res, args[i]);
+    }
+    return res;
   };
 
   api.common.altConcatLazy = (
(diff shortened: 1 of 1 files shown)


lib/functional.js
```diff
@@ -19,4 +19,89 @@ module.exports = (api) => {
     value !== undefined && value !== null ? fn(value) : defVal
   );
 
+  // Either can be used for saving either left type or right type value.
```
tshemsedinov: Description duplicated 2 times

lib/functional.js
```diff
@@ -19,4 +19,89 @@ module.exports = (api) => {
...
+  // Left type is commonly used for additional information like error.
+  //
+  function Either(
+    // Either can be used for saving either left type or right type value.
+    // Left type is commonly used for additional information like error.
+    left, // left type value
+    right // right type value
+    // Returns tuple containing left type value or null as first element
+    // and right type value or undefined as second element
+  ) {
```
tshemsedinov:
```
this.isLeft = left;
this.val = left ? left : right;
```

belochub: @tshemsedinov, why not make second line like this?
```js
this.val = left || right;
```

lib/functional.js
```diff
@@ -19,4 +19,89 @@ module.exports = (api) => {
...
+    e2 // second either value
+    // Returns result of appending two either values
+  ) => (e1[0] ? e1 : e2);
+
+  api.common.altConcat = (
+    // Alternative concatenation for either values
+    ...args // arguments for concatenation
+    // Returns result of concatenation of args
+  ) => {
+    switch (args.length) {
```
tshemsedinov: I'd prefer not using `switch` here, better `if (args.length === 0) return new Error('...');`

lib/functional.js
```diff
@@ -19,4 +19,89 @@ module.exports = (api) => {
...
+      }
+    }
+  };
+
+  api.common.altConcatLazy = (
+    // Alternative lazy concatenation for either values
+    ...args // array of functions returning either values
+    // Returns result of concatenation of either values
+  ) => {
+    switch (args.length) {
```
tshemsedinov: Use `if` instead of `switch`

lib/functional.js
tshemsedinov: @belochub sure! :+1:

DzyubSpirit:
@tshemsedinov It's seems impossible to implement `altAppend` using closures because then `altAppend` function has to be declared in both `left1`, `right1` scope and `left2`, `right2` scope. It rule can be broken after implementing `getLeft`/`getRight`.
But I think it would be useless because encapsulation is good for systems with complex internal logic but isn't same good for simple structures like Either or Point(x, y).

lib/functional.js
DzyubSpirit: @tshemsedinov, I am sorry it's my bad. I made if checking for a reason but forgot write good condition. I want that Either can save values of primitive types (not only errors) as left value. That wish needs complex checking of error.
```javascript
const [left, right] = api.common.makeEither(0); // left === 0, right === undefined
// Common checking
if (left) {
  console.log('left value');
} else {
  console.log('right value');
}
// prints 'right value'
// Complex checking
if (left !== null && left !== undefined) {
  console.log('left left');
} else {
  console.log('right value');
}
// prints 'left value'
```
That also shows hash implementation advantage. With tuple error checking has to be done every time. But with hash error checking is done once in constructor and then checking simplifies to checking `isLeft` flag

tshemsedinov:
```
this.isLeft = left !== null && left !== undefined;
this.val = this.isLeft ? left : right;
```

PR #34, opened 2017-06-13
Add tap testing to package.json

DzyubSpirit:


DzyubSpirit:
Tests fail because no tests in test folder and tap executes in command line mode. And I think it's normal because this PR is about changes for all tests and shouldn't include any specific tests.

DzyubSpirit:
Landed in fbecae4d5d1189b8d

PR #38, opened 2017-06-20
Cb extract

DzyubSpirit:


lib/misc.js
```diff
@@ -19,6 +19,20 @@ module.exports = (api) => {
...
 
+  api.common.cbExtract = (
+    // Exctracts callback and wraps it with api.common.cb
+    // callback is last argument if it's function
+    // otherwise it's api.common.falseness
+    args // arguments
+    // Returns wrapped callback
+  ) => {
+    const argsLast = args[args.length - 1];
+    if (typeof argsLast !== 'function') return api.common.falseness;
```
tshemsedinov: This code is also included in different PR. See #39 In future... that is not ok

PR #39, opened 2017-06-20
Rest left

DzyubSpirit:


lib/misc.js
```diff
@@ -19,6 +19,20 @@ module.exports = (api) => {
...
 
+  api.common.cbExtract = (
+    // Exctracts callback and wraps it with api.common.cb
+    // callback is last argument if it's function
+    // otherwise it's api.common.falseness
+    args // arguments
+    // Returns wrapped callback
+  ) => {
+    const argsLast = args[args.length - 1];
+    if (typeof argsLast !== 'function') return api.common.falseness;
```
tshemsedinov: Hint just for consistency: we use function-like syntax for `typeof()`

lib/misc.js
```diff
@@ -19,6 +19,20 @@ module.exports = (api) => {
...
   };
 
+  api.common.cbExtract = (
+    // Exctracts callback and wraps it with api.common.cb
+    // callback is last argument if it's function
+    // otherwise it's api.common.falseness
+    args // arguments
+    // Returns wrapped callback
+  ) => {
+    const argsLast = args[args.length - 1];
```
tshemsedinov: I'd like name `cb` here

lib/misc.js
```diff
@@ -89,4 +103,17 @@ module.exports = (api) => {
     return ee;
   };
 
+  api.common.restLeft = (
+    // Transforms function with (args, arg1 .. argN, callback) arguments
```
tshemsedinov: I think `(args, callback)` to `(arg1, arg2, argN, callback)`

lib/misc.js
```diff
@@ -89,4 +103,17 @@ module.exports = (api) => {
...
+    // to function with (arg1 .. argN, ...args, callback) arguments
+    fn // transforming function
+    // Returns transformed function
+  ) => (...spreadArgs) => {
+    const callback = api.common.cbExtract(spreadArgs);
+    const namedArgsCount = fn.length - 2;
```
tshemsedinov: Why `-2` ?

lib/misc.js
DzyubSpirit: @tshemsedinov I've been thinking about that name and have decided to name `argsLast` because it's last of arguments (which `restLeft` receives) for sure but it can be either callback or last of arguments (which wrapped function receives). Therefore `cb` would be false name if no callback was passed to function.
But I will try to use `cb` name.

lib/misc.js
DzyubSpirit: @tshemsedinov, I don't like using `typeof` with function-like syntax because `typeof` isn't function and it can't be saved to variable or passed to function as argument. Therefore operator-like syntax shows non-function nature of `typeof`.
But I will correct in that case because I know that conversation about code-style should be not here

lib/misc.js
DzyubSpirit: @tshemsedinov, You've described `toAsync` contract but as you say [here](https://github.com/metarhia/metasync/issues/145#issuecomment-309774531), `restLeft` is more generalized.

lib/misc.js
DzyubSpirit: @tshemsedinov, one for `callback` and one for `args`

lib/misc.js
tshemsedinov: `argsLast` means multiple, you may use `argLast`, `last`, `cb` or `callback`

DzyubSpirit: `lastArgs` means multiple, `argsLast`means args' last

lib/misc.js
```diff
@@ -19,6 +19,21 @@ module.exports = (api) => {
...
+    // Exctracts callback and wraps it with api.common.cb
+    // callback is last argument if it's function
+    // otherwise it's api.common.falseness
+    args // arguments
+    // Returns wrapped callback
+  ) => {
+    const lastArg = args[args.length - 1];
+    if (typeof(lastArg) !== 'function') return api.common.falseness;
+
+    const cb = lastArg;
```
tshemsedinov: `const cb = api.common.cb(lastArg);`

DzyubSpirit:
Landed in efa9e92ea03be0d4c3b, 02e6276bca8

PR #44, opened 2017-06-23
Require either

DzyubSpirit:


DzyubSpirit:
Landed in 755bdbe549ff18c546, 4d9fe9c2886ad78c

PR #52, opened 2017-07-03
Improvements to lib/functional

IvanProdaiko94:
- remove bind from `curry`;
- `replicate` function rewritten;
- do not create array when it is not needed.

### DIFF
--- a/lib/functional.js
+++ b/lib/functional.js
@@ -39,7 +39,7 @@ module.exports = (api) => {
     // Element with index i of resulting array is array with
     // elements with index i from input arrays.
   ) => {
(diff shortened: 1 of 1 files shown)


tshemsedinov:
@IvanProdaiko94 here https://github.com/metarhia/common/commit/17719c13c2ef489d03808638d9e3ef7f2d3821b6 @DzyubSpirit did it but I think such curry with lambdas can be done without additional utilities, what we really need is a `fullCurry`. For compatibility with http://ramdajs.com/docs/#curry we can name it `curry`

tshemsedinov:
Landed in 56f8043

PR #85, opened 2017-08-12
Remove duplicate code

DzyubSpirit:


lib/data.js
```diff
@@ -74,19 +102,7 @@ function duplicateArray(arr, references) {
...
-      array[i] = references.get(val);
-    } else  if (typeof(val) !== 'object' || val === null) {
-      array[i] = val;
-    } else if (Array.isArray(val)) {
-      array[i] = duplicateArray(val, references);
-    } else {
-      array[i] = duplicateObject(val, references);
-    }
-  }
+  duplicateFields(references, arr, array, range(0, size - 1));
```
tshemsedinov: I don't think generating new array using `range` and then iteration through this array will be faster than `for-loop` through integer indexes of array.

lib/data.js
```diff
@@ -23,39 +25,65 @@ const clone = (
...
-      array[i] = cloneObject(val);
+      dstObj[fieldName] = transformArrObj(val, duplicateArray, duplicateObject);
     }
   }
+};
+
+function cloneArray(arr) {
+  const size = arr.length;
+  const array = new Array(size);
+  cloneFields(arr, array, range(0, size - 1));
```
tshemsedinov: See comments for `duplicateArray`

lib/data.js
```diff
@@ -23,39 +25,65 @@ const clone = (
...
-  const size = arr.length;
-  const array = new Array(size);
-  let i, val;
-  for (i = 0; i < size; i++) {
-    val = arr[i];
-    if (typeof(val) !== 'object' || val === null) {
-      array[i] = val;
-    } else if (Array.isArray(val)) {
-      array[i] = cloneArray(val);
+const transformArrObj = (
```
tshemsedinov: Bad naming, it's not obvious what `transform` means. Whole name `transformArrObj` is not clear at all.

DzyubSpirit: Is `handle` suitable here? Could whole name be `handleArrOrObj`?

PR #88, opened 2017-08-13
Implement deprecate() to mark methods as deprecated

tshemsedinov:


tshemsedinov:
Landed in 26be701

tshemsedinov:
We need

tshemsedinov:
Landed in 26be701 and 0d404d4

PR #97, opened 2017-09-02
Remove duplicate code using loop

DzyubSpirit:
PR with looping without making array of keys. Created to compare speed with #85 using `majian`


tshemsedinov:
@bugagashenkj can you please rebase this on master then review PR with @lundibundi

PR #98, opened 2017-09-19
Fix curry test

tshemsedinov:


DzyubSpirit:
@tshemsedinov, It isn't really fix of test because it's intendent to use additional arguments as arguments to curried function. However it can be a new useful test case

test/curry.js
```diff
@@ -5,7 +5,7 @@ const common = require('..');
 
 tap.test('curry(f)(1)(2)(3)', (test) => {
```
tshemsedinov: @DzyubSpirit you wrote this case in test title but it differed in code

tshemsedinov:
Hard to rebase

PR #101, opened 2017-09-19
Reimplemented curry alternatively

tshemsedinov:


DzyubSpirit:
@tshemsedinov, That implementation creates closure every time curried function is called. `curry(fn, 1)(2)(3)` creates 3 closures as I understand. Previous implementation always returns `curryMore` function  (in `curryUntil`) created only in first call of `curry`. 
Solution gathering advantages of both implementation is something like that, I think:
```javascript
const curry = (
    // Curry function with given arguments
    fn, // function
    ...par // array, arguments
    // Returns: function, curried
  ) => {
  if (fn.length <= par.length) {
    return fn(...par);
  }

  const curried = (...args) => {
    par = par.concat(args);
    return fn.length > par.length ? curried : fn(...par);
  };
  return curried();
};
```

aqrln:
This looks good (although I have not tested it) and is a considerable improvement over the current version in `master`, which seems to be prone to a similar bug as in https://github.com/metarhia/common/pull/199#issuecomment-425905362.

lundibundi:
While we are at it I've also changed the comments style.

test/curry.js
```diff
@@ -44,3 +44,25 @@ metatests.test('curry(f)(1, 2, 3)', (test) => {
...
+metatests.testSync('multiple curry of sum(x, y, z)', (test) => {
+  const sum = (x, y, z) => x + y + z;
+  const sumCurry = common.curry(sum);
+  const addOneTwo = sumCurry(1, 2);
+  const addTwoThree = sumCurry(2, 3);
+  test.strictSame(addOneTwo(10), 13);
+  test.strictSame(addOneTwo(20), 23);
+  test.strictSame(addTwoThree(10), 15);
+  test.strictSame(addTwoThree(20), 25);
+});
```
aqrln: Maybe also add some tests for `x => x` and `() => x`?

lundibundi: Sounds good.

test/curry.js
```diff
@@ -66,3 +66,20 @@ metatests.testSync('multiple curry of sum(x, y, z)', (test) => {
...
+  test.strictSame(idCurry(20), 20);
+
+  test.strictSame(common.curry(id, 10), 10);
+});
+
+metatests.testSync('curry of unit', (test) => {
+  const unit = () => 42;
+  const unitCurry = common.curry(unit);
+  test.strictSame(unitCurry(), 42);
+  test.strictSame(unitCurry(10), 42);
```
aqrln: I don't think this test and the next one make much sense theoretically, it's just how JavaScript inherently works and not a part of the spec of the tested function.

That said, I'm not strongly opposed to it, so if there already are some tests that call a curried function with more arguments than it's supposed to have, let's keep these too for consistency. If there are none yet, I'm -0 on adding them.

aqrln: Also, if we were to add these, I'd probably favor an explicit unit value, like `undefined`, instead of `10`. This, at least, can have some far-fetched theoretical justification of why it makes sense.

test/curry.js
```diff
@@ -44,3 +44,52 @@ metatests.test('curry(f)(1, 2, 3)', (test) => {
...
+  test.strictSame(common.curry(id, 10), 10);
+});
+
+metatests.testSync('curry of unit', (test) => {
+  const unit = () => 42;
+  const unitCurry = common.curry(unit);
+  test.strictSame(unitCurry(), 42);
+});
+
+metatests.testSync('reduntant args must be ignored', (test) => {
```
aqrln: typo: redundant

PR #108, opened 2017-09-28
Use strings instead of regexp

bugagashenkj:
Replaced regexp by string in file units.js

### DIFF
--- a/lib/units.js
+++ b/lib/units.js
@@ -1,10 +1,10 @@
 'use strict';
 
 const DURATION_UNITS = {
-  days:    { rx: /(\d+)\s*d/, mul: 86400 },
-  hours:   { rx: /(\d+)\s*h/, mul: 3600 },
-  minutes: { rx: /(\d+)\s*m/, mul: 60 },
-  seconds: { rx: /(\d+)\s*s/, mul: 1 }
+  d: 86400, // days
+  h:  3600, // hours
+  m:    60, // minutes
+  s:     1, // seconds
 };
 
 const duration = (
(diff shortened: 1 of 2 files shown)


aqrln:
Hi and thanks for your contribution! As you might have noticed, these changes [break 17 tests](https://travis-ci.org/metarhia/common/jobs/280872940#L1059). Since you are a first-time GitHub contributor, I believe it is worth noting that you can push new commits to your branch, and they will automatically appear right here in this PR. And, for future contributions, please don't forget to run `npm test` before committing ;)
Feel free to ask any questions in this thread if you need some help.

Kowalski0805:
Hello, I fixed sizeToBytes function, everything is okay now :)

lib/units.js
```diff
@@ -7,6 +7,18 @@ const DURATION_UNITS = {
   s:     1, // seconds
 };
 
+const find = (
+  // Extract substring between prefix and suffix
```
tshemsedinov: Where are prefix and suffix?

lib/units.js
```diff
@@ -59,16 +71,16 @@ const sizeToBytes = (
...
-  size = size.toUpperCase();
+  size = size.toLowerCase();
   let result = 0;
   if (typeof(size) === 'string') {
     let key, unit, match;
     let found = false;
     for (key in UNIT_SIZES) {
       unit = UNIT_SIZES[key];
-      match = size.match(unit.rx);
+      match = find(size, ' ' + key);
```
tshemsedinov: `match` should be renamed too because we don't store regexp match result in it

lib/units.js
```diff
@@ -16,11 +28,11 @@ const duration = (
   if (typeof(s) === 'number') return s;
   if (typeof(s) !== 'string') return 0;
   let result = 0;
-  let key, unit, match;
+  let key, unit;
   for (key in DURATION_UNITS) {
     unit = DURATION_UNITS[key];
-    match = s.match(unit.rx);
-    if (match) result += parseInt(match[1], 10) * unit.mul;
+    const subst = parseInt(find(s, key));
```
tshemsedinov: Move `subst` declaration outside of loop and make it `let`

lib/units.js
```diff
@@ -7,6 +7,18 @@ const DURATION_UNITS = {
...
 
+const find = (
+  // Extract substring between prefix and suffix
+  s, // string, source
+  key // string
+  // Returns: string
+) => {
+  const i = s.indexOf(key);
+  if (i === -1) return '';
+  s = s.substring(0, i).split(' ');
```
tshemsedinov: Do not change `s` because it's function argument

tshemsedinov:
I'll land this because it's better then current version `valueByUnit` should be optimized to avoid string parsing for each unit.

tshemsedinov:
Landed in c3a43a2, 3717c09 and efb804c

PR #120, opened 2018-02-10
Fix and refactor setByPath function

belochub:


belochub:
Landed in https://github.com/metarhia/common/commit/6b335a0f54d8858bdbb233b503c684d42c0cb530.

Issue #122, opened 2018-02-27
Add map,filter, etc. for iterators


nechaido:
I propose to add an `Iterator` class that has `map`, `filter`, etc. functions. Example syntax:
```js
const set = new Set([1, 2, 3, 4]);

const doubledArray = new Iterator(set)
  .map(item => item * 2)
  .fetch();

const ten = new Iterator(set)
  .reduce((previous, current) => previous + current, 0);

new Iterator(set).forEach(item => console.log(item));
```

aqrln:
Love the idea, but shouldn't it be a part of `metasync` instead though?

nechaido:
@aqrln `metasync` is for async functionality. I will to add `AsyncIterator` for `metasync` as soon as @tshemsedinov  approves it.

PR #123, opened 2018-02-28
Add Iterator implementation

nechaido:
Closes: #122

### DIFF
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -1,0 +1,304 @@
+/* eslint-disable no-use-before-define */
+
+'use strict';
+
+const toIterator = base => {
+  if (!base[Symbol.iterator]) {
+    throw new TypeError('Base is not Iterable');
+  }
+  return base[Symbol.iterator]();
+};
+
+class Iterator {
+  constructor(base) {
+    this.base = toIterator(base);
+  }
+
+  [Symbol.iterator]() {
+    return this;
+  }
+
+  next() {
+    return this.base.next();
+  }
+
+  each(fn, thisArg) {
+    this.forEach(fn, thisArg);
+  }
+
+  forEach(fn, thisArg) {
+    for (const value of this) {
+      fn.call(thisArg, value);
+    }
+  }
+
+  every(predicate, thisArg) {
+    for (const value of this) {
+      if (!predicate.call(thisArg, value)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  find(predicate, thisArg) {
+    for (const value of this) {
+      if (predicate.call(thisArg, value)) {
+        return value;
+      }
+    }
+  }
+
+  includes(element) {
+    for (const value of this) {
+      if (value === element || (isNaN(value) && isNaN(element))) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  reduce(reducer, initialValue) {
+    let result = initialValue;
+
+    if (result === undefined) {
+      const next = this.next();
+      if (next.done) {
+        throw new TypeError(
+          'Reduce of consumed iterator with no initial value'
+        );
+      }
+      result = next.value;
+    }
+
+    for (const value of this) {
+      result = reducer(result, value);
+    }
+    return result;
+  }
+
+  some(predicate, thisArg) {
+    for (const value of this) {
+      if (predicate.call(thisArg, value)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  collectTo(CollectionClass) {
+    return new CollectionClass(this);
+  }
+
+  collectWith(obj, collector) {
+    this.forEach(element => collector(obj, element));
+  }
+
+  toArray() {
+    return Array.from(this);
+  }
+
+  map(mapper, thisArg) {
+    return new MapIterator(this, mapper, thisArg);
+  }
+
+  filter(predicate, thisArg) {
+    return new FilterIterator(this, predicate, thisArg);
+  }
+
+  flat(depth = 1) {
+    return new FlatIterator(this, depth);
+  }
+
+  flatMap(mapper, thisArg) {
+    return new FlatMapIterator(this, mapper, thisArg);
+  }
+
+  zip(...iterators) {
+    return new ZipIterator(this, iterators);
+  }
+
+  join(...iterators) {
+    return new JoinIterator(this, iterators);
+  }
+
+  take(amount) {
+    return new TakeIterator(this, amount);
+  }
+
+  skip(amount) {
+    for (let i = 0; i < amount; i++) {
+      this.next();
+    }
+    return this;
+  }
+}
+
+class MapIterator extends Iterator {
+  constructor(base, mapper, thisArg) {
+    super(base);
+    this.mapper = mapper;
+    this.thisArg = thisArg;
+  }
+
+  next() {
+    const { done, value } = this.base.next();
+    return {
+      done,
+      value: done ? undefined : this.mapper.call(this.thisArg, value),
+    };
+  }
+}
+
+class FilterIterator extends Iterator {
+  constructor(base, predicate, thisArg) {
+    super(base);
+    this.predicate = predicate;
+    this.thisArg = thisArg;
+  }
+
+  next() {
+    for (const value of this.base) {
+      if (this.predicate.call(this.thisArg, value)) {
+        return { done: false, value };
+      }
+    }
+    return { done: true, value: undefined };
+  }
+}
+
+class FlatIterator extends Iterator {
+  constructor(base, depth) {
+    super(base);
+    this.currentDepth = 0;
+    this.stack = new Array(depth + 1);
+    this.stack[0] = base;
+  }
+
+  next() {
+    while (this.currentDepth >= 0) {
+      const top = this.stack[this.currentDepth];
+      const next = top.next();
+
+      if (next.done) {
+        this.stack[this.currentDepth] = null;
+        this.currentDepth--;
+        continue;
+      }
+
+      if (
+        this.currentDepth === this.stack.length - 1 ||
+        !next.value[Symbol.iterator]
+      ) {
+        return next;
+      }
+
+      this.stack[++this.currentDepth] = next.value[Symbol.iterator]();
+    }
+
+    return { done: true, value: undefined };
+  }
+}
+
+class FlatMapIterator extends Iterator {
+  constructor(base, mapper, thisArg) {
+    super(base);
+    this.mapper = mapper;
+    this.thisArg = thisArg;
+    this.currentIterator = null;
+  }
+
+  next() {
+    if (!this.currentIterator) {
+      const next = this.base.next();
+      if (next.done) {
+        return next;
+      }
+
+      const value = this.mapper.call(this.thisArg, next.value);
+      if (!value[Symbol.iterator]) {
+        return { done: false, value };
+      }
+
+      this.currentIterator = toIterator(value);
+    }
+
+    const next = this.currentIterator.next();
+
+    if (next.done) {
+      this.currentIterator = null;
+      return this.next();
+    }
+    return next;
+  }
+}
+
+class TakeIterator extends Iterator {
+  constructor(base, amount) {
+    super(base);
+    this.amount = amount;
+    this.iterated = 0;
+  }
+
+  next() {
+    this.iterated++;
+    if (this.iterated <= this.amount) {
+      return this.base.next();
+    }
+    return { done: true, value: undefined };
+  }
+}
+
+class ZipIterator extends Iterator {
+  constructor(base, iterators) {
+    super(base);
+    this.iterators = iterators.map(toIterator);
+  }
+
+  next() {
+    const result = [];
+
+    const next = this.base.next();
+    if (next.done) {
+      return next;
+    }
+    result.push(next.value);
+
+    for (const iterator of this.iterators) {
+      const next = iterator.next();
+      if (next.done) {
+        return next;
+      }
+      result.push(next.value);
+    }
+    return { done: false, value: result };
+  }
+}
+
+class JoinIterator extends Iterator {
+  constructor(base, iterators) {
+    super(base);
+    this.currentIterator = base;
+    this.iterators = iterators.map(toIterator)[Symbol.iterator]();
+  }
+
+  next() {
+    const next = this.currentIterator.next();
+    if (!next.done) {
+      return next;
+    }
+    const iterator = this.iterators.next();
+    if (iterator.done) {
+      return iterator;
+    }
+    this.currentIterator = iterator.value;
+    return this.next();
+  }
+}
+
+const iter = base => new Iterator(base);
+
+module.exports = {
+  Iterator,
+  iter,
+};
--- a/test/iterator.js
+++ b/test/iterator.js
@@ -1,0 +1,354 @@
+'use strict';
+
+const { Iterator, iter } = api.common;
+
+const array = [1, 2, 3, 4];
+
+api.metatests.test('new Iterator() on non Iterable', test => {
+  test.throws(() => {
+    new Iterator(2);
+  }, new TypeError('Base is not Iterable'));
+  test.end();
+});
+
+api.metatests.test('iter returns an Iterator', test => {
+  const iterator = iter(array);
+  test.assert(iterator instanceof Iterator);
+  test.end();
+});
+
+api.metatests.test('Iterator is Iterable', test => {
+  const iterator = iter(array);
+  let sum = 0;
+  for (const value of iterator) {
+    sum += value;
+  }
+
+  test.strictSame(sum, 10);
+  test.end();
+});
+
+api.metatests.test('Iterator.each', test => {
+  const iterator = iter(array);
+  let sum = 0;
+  iterator.each(value => {
+    sum += value;
+  });
+
+  test.strictSame(sum, 10);
+  test.end();
+});
+
+api.metatests.test('Iterator.forEach', test => {
+  const iterator = iter(array);
+  let sum = 0;
+  iterator.forEach(value => {
+    sum += value;
+  });
+
+  test.strictSame(sum, 10);
+  test.end();
+});
+
+api.metatests.test('Iterator.forEach with thisArg ', test => {
+  const iterator = iter(array);
+  const obj = {
+    sum: 0,
+    fn(value) {
+      this.sum += value;
+    },
+  };
+
+  iterator.forEach(obj.fn, obj);
+
+  test.strictSame(obj.sum, 10);
+  test.end();
+});
+
+api.metatests.test('Iterator.reduce', test => {
+  test.strictSame(iter(array).reduce((acc, current) => acc + current, 0), 10);
+  test.end();
+});
+
+api.metatests.test('Iterator.reduce with no initialValue', test => {
+  test.strictSame(iter(array).reduce((acc, current) => acc + current), 10);
+  test.end();
+});
+
+api.metatests.test(
+  'Iterator.reduce with no initialValue on consumed iterator',
+  test => {
+    const iterator = iter(array);
+    test.throws(() => {
+      iterator.reduce(() => {});
+      iterator.reduce((acc, current) => acc + current);
+    }, new TypeError('Reduce of consumed iterator with no initial value'));
+
+    test.end();
+  }
+);
+
+api.metatests.test('Iterator.map', test => {
+  test.strictSame(
+    iter(array)
+      .map(value => value * 2)
+      .toArray(),
+    [2, 4, 6, 8]
+  );
+  test.end();
+});
+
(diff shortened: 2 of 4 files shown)


nechaido:
@tshemsedinov this is WIP I am open to suggestions and there no tests, I will add them later.

lundibundi:
@tshemsedinov can we proceed with this one? It's quite useful to have `forEach`, `some`, `any` etc on any Iterator. Though tests should be ported to metatests.

lib/iterator.js
```diff
@@ -0,0 +1,115 @@
+'use strict';
+
+class Iterator {
+  constructor(base) {
+    if (base[Symbol.iterator]) {
+      this.base = base[Symbol.iterator]();
+    } else if (base.next) {
+      this.base = base;
+    } else {
+      throw new TypeError('Base is neither Iterator nor Itarable');
```
lundibundi: s/Itarable/Iterable

lib/iterator.js
```diff
@@ -0,0 +1,115 @@
...
+  next() {
+    const { done, value } = this.base.next();
+    return { done, value: done ? undefined : this.mapper(value) };
+  }
+}
+
+class FilterIterator extends Iterator {
+  constructor(base, reducer) {
+    super(base);
+
```
lundibundi: Redundant blank

tshemsedinov:
Rebase needed only for single file, even for single line

nechaido:
@tshemsedinov PTAL at tests, is it the right way to write tests with metatest?

belochub:
I don't like the fact that all of the methods on these `Iterator`s have a behavior different to the built-in method's behavior in terms of `this` reference usage.

Built-in methods allow to provide `thisArg` as the last argument to the method, which then has an effect on callback calls [as described on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach#Description) (on an example of `Map.prototype.forEach`):

>If a `thisArg` parameter is provided to `forEach`, it will be passed to `callback` when invoked, for use as its `this` value.  Otherwise, the value `undefined` will be passed for use as its `this` value.  The `this` value ultimately observable by `callback` is determined according to the usual rules for determining the `this` seen by a function.

And if you take a look at all other similar built-in methods, they all seem to have this behavior.

@nechaido, can you, please, change this `Iterator` implementation to have similar behavior in terms of `this` usage?

nechaido:
@belochub I would prefer not to, but if @tshemsedinov approves your request I will implement the desired behaviour.

tshemsedinov:
@nechaido I'd agree with @belochub, it will be good to be contact compatible if it have no significant performance influence

nechaido:
@tshemsedinov what do you think about:
```js
  collect(Class) {
    return new Class(...this);
  }
```

tshemsedinov:
`.toCollect(CollectionClass)`

aqrln:
How about `.collectTo(CollectionClass)`?

nechaido:
@tshemsedinov please take a look at Trevis, this looks to me like a problem with the metarhia/metatests itself.
It appears that on node v6 it places every test in a queue and starts executing them only after every test is placed in the queue, while on node v7 and higher test are executed at the time api.metasync.test() is called.

tshemsedinov:
@nechaido it works for me with node 6.14.1 locally. I'll try to update metatests to my latest patch

lib/iterator.js
```diff
@@ -0,0 +1,132 @@
...
+  find(predicate, thisArg) {
+    for (const value of this) {
+      if (predicate.call(thisArg, value)) {
+        return value;
+      }
+    }
+  }
+
+  includes(element) {
```
lundibundi: Could we add [`fromIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Parameters)?

lib/iterator.js
```diff
@@ -0,0 +1,132 @@
...
+    // eslint-disable-next-line no-use-before-define
+    return new MapIterator(this, mapper, thisArg);
+  }
+
+  filter(predicate, thisArg) {
+    // eslint-disable-next-line no-use-before-define
+    return new FilterIterator(this, predicate, thisArg);
+  }
+
+  each(predicate, thisArg) {
```
lundibundi: To be compliant with JS Array s/each/every?

lib/iterator.js
```diff
@@ -0,0 +1,132 @@
...
+    for (const value of this) {
+      if (predicate.call(thisArg, value)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  collectTo(CollectionClass) {
+    return new CollectionClass(...this);
```
lundibundi: Can the spread be omitted (this is iterable)?

test/iterator.js
```diff
@@ -0,0 +1,239 @@
+'use strict';
+
+const { Iterator } = api.common;
+
+const array = [1, 2, 3, 4];
+
+const test = (name, test) => api.metatests.test(name, test);
+
+test('Iterator is iterable', test => {
+  const iterator = new Iterator(array);
```
lundibundi: I assume we don't have, `beforeEach`, `afterEach` functionality in metatests? =(

test/iterator.js
```diff
@@ -0,0 +1,239 @@
...
+  }, new TypeError('reduce of consumed iterator with no initial value'));
+
+  test.end();
+});
+
+test('Iterator.map', test => {
+  const iterator = new Iterator(array);
+  const sum = iterator
+    .map(value => value * 2)
+    .reduce((previous, current) => previous + current, 0);
```
lundibundi: s/previous/acc? Just a nit though.

test/iterator.js
tshemsedinov: We have no

lib/iterator.js
tshemsedinov: I'd like :+1:

nechaido: An index doesn't make sense for `Map`, `Set` or `Iterable` in general for what it is worth.

tshemsedinov:
I'd like to have `forEach` synonym `each`

lib/iterator.js
lundibundi: That's true, the what do you think about .take and .skip methods?

test/iterator.js
aqrln: Or `accumulator` fully

nechaido:
I would prefer not to add any new functionality at his point this PR is already too big.

lundibundi:
@tshemsedinov imo `each` maybe be confusing with `every` in JS and `all` in other languages.
Though I agree with @nechaido that we should land this one and add more functionality in follow up PRs.

nechaido:
@aqrln ping.

lib/iterator.js
```diff
@@ -0,0 +1,312 @@
...
+    }
+    return { done: true, value: undefined };
+  }
+}
+
+class FlatIterator extends Iterator {
+  constructor(base, depth) {
+    super(base);
+    this.depth = depth;
+    this.stack = [base];
```
belochub: You know the maximum stack size at the time of creation, so you may as well just create a fixed size array here, to avoid additional memory allocations.

lib/iterator.js
```diff
@@ -0,0 +1,312 @@
...
+    // eslint-disable-next-line no-use-before-define
+    return new FilterIterator(this, predicate, thisArg);
+  }
+
+  flat(depth = 1) {
+    // eslint-disable-next-line no-use-before-define
+    return new FlatIterator(this, depth);
+  }
+
+  flatMap(mapper, thisArg) {
```
belochub: This function throws on this basic valid example:
```js
common.iter([1, 2, 3]).flatMap(x => x*2).toArray()
TypeError: Base is neither Iterator nor Iterable
    at toIterator (common/lib/iterator.js:9:11)
    at new Iterator (common/lib/iterator.js:15:17)
    at FlatMapIterator.next (common/lib/iterator.js:230:30)
    at Function.from (native)
    at FlatMapIterator.toArray (common/lib/iterator.js:99:18)
```

lib/iterator.js
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  zip(...iterators) {
+    // eslint-disable-next-line no-use-before-define
+    return new ZipIterator(this, iterators);
+  }
+
+  join(...iterators) {
+    // eslint-disable-next-line no-use-before-define
+    return new JoinIterator(this, new Iterator(iterators));
```
belochub: Why do you need to create `new Iterator` from `iterators` array here?

lib/iterator.js
```diff
@@ -0,0 +1,312 @@
...
+    const next = this.currentIterator.next();
+    if (!next.done) {
+      return next;
+    }
+    const { done, value: iterator } = this.iterators.next();
+    if (done) {
+      return next;
+    }
+    this.currentIterator = iterator;
+    return this.next();
```
belochub: Can you do this without recursion?
This seems to lead to getting an error with maximum call stack size exceedance.

test/iterator.js
```diff
@@ -0,0 +1,359 @@
...
+  test.strictSame(
+    iter(array)
+      .flatMap(element => [element, element])
+      .toArray(),
+    result
+  );
+  test.end();
+});
+
+api.metatests.test('Iterator.flatMap', test => {
```
belochub: It should probably be called `'Iterator.flatMap with thisArg'`.

lib/iterator.js
nechaido: So that I don't need to shift an array or store an index.

lib/iterator.js
nechaido: I wouldn`t like to do so, because then I'll need to store the current index. Also, this looks like premature optimisations. You can change that in a follow-up PR.

lib/iterator.js
lundibundi: iterators[Symbol.iterator] or toIterator(iterators)?

lib/iterator.js
nechaido: This looks like premature optimisations. You can change that in a follow-up PR. You'll need to join a lot of empty iterators

lib/iterator.js
belochub: You don't need to store current index, you can use `depth` field for that.

lib/iterator.js
belochub: @nechaido, but you are already converting this array to an iterator [here](https://github.com/metarhia/common/pull/123/files/28acd54c78c568489f2b40df9fad8fd0b698edd3#diff-d217736e9b27c173b1b076c6fbd6416aR290).

nechaido: And what if iterators is an Iterator itself?

test/iterator.js
nechaido: Fixed.

lib/iterator.js
belochub: @nechaido, `iterators` is an array here.

lib/iterator.js
nechaido: I need to know both desired depth and current depth, that's why I'll need an additional index.

nechaido: Yea, I see now, I will not need a `depth` variable as desired depth can be inferred from `this.stack.length`.

lib/iterator.js
nechaido: Fixed.

lib/iterator.js
nechaido: Implemented `skip(amount)` and `take(amount)`.

lib/iterator.js
```diff
@@ -0,0 +1,317 @@
...
+      if (next.value[Symbol.iterator]) {
+        this.stack[++this.depth] = next.value[Symbol.iterator]();
+        continue;
+      }
+
+      if (!next.value.next) {
+        return next;
+      }
+
+      this.stack[++this.depth] = next.value[Symbol.iterator]();
```
belochub: Typo here, this line should be
```js
this.stack[++this.depth] = next.value;
```

lib/iterator.js
```diff
@@ -0,0 +1,317 @@
...
+
+  next() {
+    if (!this.currentIterator) {
+      const next = this.base.next();
+      if (next.done) {
+        return next;
+      }
+
+      const value = this.mapper.call(this.thisArg, next.value);
+      if (!value[Symbol.iterator] && !value.next) {
```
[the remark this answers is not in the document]
aqrln: But frankly I'd just drop support for iterators that are not iterables, because I can't see any valid use case for them, but I do perceive non-obvious bugs in application code. Can you please give an example when they might be useful?

lib/iterator.js
```diff
@@ -0,0 +1,317 @@
...
+
+class TakeIterator extends Iterator {
+  constructor(base, amount) {
+    super(base);
+    this.amount = amount;
+    this.iterated = 0;
+  }
+
+  next() {
+    if (this.iterated++ < this.amount) {
```
aqrln: Not being strongly opposed to this as it is, but I'd rather split this into two statements.

lib/iterator.js
```diff
@@ -0,0 +1,317 @@
...
+      }
+    }
+    return false;
+  }
+
+  collectTo(CollectionClass) {
+    return new CollectionClass(this);
+  }
+
+  collectWith(obj, collector) {
```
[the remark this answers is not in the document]
nechaido: @lundibundi you were the one who requested this particular implementation.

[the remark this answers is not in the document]
lundibundi: Yeah, it's just `forEach` with a bit of state. I suggested it to be able to do
`iter(...).collectWith(new Struct(), (el, obj) => obj.add(el))`
Instead of having 
```
const obj = new Struct()
iter(...).forEach(el => obj.add(el))
```
Or
```
iter(...).reduce((obj, el) => (obj.add(el), obj), new Struct())
```

lib/iterator.js
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  next() {
+    const next = this.currentIterator.next();
+    if (!next.done) {
+      return next;
+    }
+    const { done, value: iterator } = this.iterators.next();
+    if (done) {
+      return next;
```
aqrln: It's a bit counterintuitive and looks like a typo. Maybe add a comment that `next` is always `{ done: true, value: undefined }` at this point?

test/iterator.js
```diff
@@ -0,0 +1,367 @@
...
+
+    test.end();
+  }
+);
+
+api.metatests.test('Iterator.map', test => {
+  test.strictSame(
+    iter(array)
+      .map(value => value * 2)
+      .reduce((previous, current) => previous + current),
```
aqrln: Can `metatests` shallow-compare arrays for equality? Using `reduce()` in a unit test for `map()` doesn't sound right, IMHO.

test/iterator.js
```diff
@@ -0,0 +1,367 @@
...
+    20
+  );
+  test.end();
+});
+
+api.metatests.test('Iterator.filter', test => {
+  test.strictSame(
+    iter(array)
+      .filter(value => !(value % 2))
+      .reduce((previous, current) => previous + current),
```
aqrln: ditto

test/iterator.js
```diff
@@ -0,0 +1,367 @@
...
+    multiplier: 2,
+    mapper(value) {
+      return value * this.multiplier;
+    },
+  };
+
+  test.strictSame(
+    iter(array)
+      .map(obj.mapper, obj)
+      .reduce((previous, current) => previous + current),
```
aqrln: ditto

test/iterator.js
```diff
@@ -0,0 +1,367 @@
...
+    divider: 2,
+    predicate(value) {
+      return !(value % this.divider);
+    },
+  };
+
+  test.strictSame(
+    iter(array)
+      .filter(obj.predicate, obj)
+      .reduce((previous, current) => previous + current),
```
aqrln: ditto

lib/iterator.js
```diff
@@ -0,0 +1,317 @@
...
+    }
+    return { done: true, value: undefined };
+  }
+}
+
+class FlatIterator extends Iterator {
+  constructor(base, depth) {
+    super(base);
+    this.depth = 0;
+    this.stack = new Array(depth + 1);
```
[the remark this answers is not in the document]
lundibundi: Oh well, that's how it was before, lol =).

Also, I'd not be so sure about `new Array(length)` (as simple microbenchmarks show that preallocation wins by a great margin) though it indeed has to be tested on a per-case basis.

test/iterator.js
```diff
@@ -0,0 +1,367 @@
...
+
+api.metatests.test('Iterator.toArray must convert to array', test => {
+  const iterator = new Iterator(array);
+  const arr = iterator.toArray();
+  test.strictSame(arr, array);
+  test.end();
+});
+
+api.metatests.test(
+  'Iterator.collectWith must collect in an provided object',
```
aqrln: typo: s/`in an provided object`/`to a provided object`

lib/iterator.js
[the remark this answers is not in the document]
belochub: @nechaido I'd like for it to be benchmarked if possible before we make the decision to pick one or the other implementation since I am quite sceptical about the performance benefits of *not* using `new Array(length)` and access by indices (when we have the maximal array length known at the start) compared to creating an empty array and using `Array.prototype.push()` and `Array.prototype.pop()` to add/remove the elements.

lib/iterator.js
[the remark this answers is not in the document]
aqrln: @lundibundi in your particular example `Struct` constructor would probably accept an optional iterable to support `new Struct(elementsArray)`, so you'd be able to do `iter(elementsArray).collectTo(Struct)`, fwiw. That said, I understand that it's not always possible generally (e.g., you might need to collect data to a plain object). In this case, what exactly do find unergonomic in `reduce()`? Just the order of arguments or am I missing something?

lib/iterator.js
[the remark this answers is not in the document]
aqrln: @lundibundi I wasn't aware it was the previous implementation, hence a lot of comments, duh.

> Also, I'd not be so sure about new Array(length) (as simple microbenchmarks show that preallocation wins by a great margin)

If that hasn't changed over the last year, it doesn't: <https://v8project.blogspot.com/2017/09/elements-kinds-in-v8.html>.

lib/iterator.js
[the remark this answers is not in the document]
lundibundi: Yeah, basically order and the need to return the object every time even though it doesn't change. Also given the simplicity of implementation I suggested to just have it.
(Yeah, I assumed that it doesn't have such constructor/additional processing needed)

lib/iterator.js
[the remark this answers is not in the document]
aqrln: Keep in mind that a benchmark must push and pop items and access the last one repeatedly to emulate this particular use case.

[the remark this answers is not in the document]
nechaido: @aqrln @belochub I propose to rename `this.depth` to `this.currentDepth` as it may indeed be counterintuitive and leave the rest of the code as it is. Whenever someone proposes significant benchmarks that state that `[]` is faster or even equal - I'd be glad refactor this class.
Please leave your response ASAP, I would like to finally get over this PR.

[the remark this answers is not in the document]
aqrln: @nechaido sounds good to me

lib/iterator.js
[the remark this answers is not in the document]
aqrln: @lundibundi yeah, I overlooked the fact that you don't need to return the same object you mutate imperatively anyway with `collectWith`, thanks. Should we maybe keep the reducer arguments order the same though?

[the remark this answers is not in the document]
lundibundi: No strong opinion, though I assume they are this way in reduce due to the optionality of initialItem and here we have it as required. Also I kind of like when fuction comes last as it's clearer imo. Though if you or @nechaido want to keep same order as in reduce that's fine with me too.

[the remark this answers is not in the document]
aqrln: @lundibundi I agree with this. I was talking about the reducer signature, not the method one.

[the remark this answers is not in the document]
nechaido: Changed.

lib/iterator.js
```diff
@@ -0,0 +1,304 @@
...
+  next() {
+    const result = [];
+
+    const next = this.base.next();
+    if (next.done) {
+      return next;
+    }
+    result.push(next.value);
+
+    for (const iterator of this.iterators) {
```
lundibundi: In a follow up PR: update this to have `finished` flag to avoid going every time through this loop up to the iterator that is `done`.

tshemsedinov:
<details>
 <summary>LIFAS</summary>
Land it for Allah's sake!
</details>

belochub:
Landed in https://github.com/metarhia/common/commit/03f102e312d8f7ffabdc42b06fc918671afc3175.

PR #124, opened 2018-02-28
Add more secure pseudo-random number generator

belochub:
Implement function with same behavior as `Math.random()`, using `crypto.randomBytes` as value generator.


tshemsedinov:
Landed in 2e16215

PR #125, opened 2018-03-12
Optimize secure pseudo-random number generation

belochub:
Speed up secure pseudo-random number generation by prefetching the random bytes.


belochub:
Landed in https://github.com/metarhia/common/commit/db3c9866e8dada76f6d141c632da8fb2a69d355f.

PR #126, opened 2018-03-12
Fix value distribution in `generateKey` function

belochub:
`generateKey` function now returns normally distributed values.


belochub:
Landed in https://github.com/metarhia/common/commit/457dbfadd791365922e7b9402d6d4edf753faf61.

PR #127, opened 2018-03-12
 Optimize GUID generation

belochub:
GUID generation will now ask only for 16 bytes random bytes instead of 128, and will also take advantage of prefetching.

This PR is based on https://github.com/metarhia/common/pull/125 and must be landed after it.


belochub:
Landed in https://github.com/metarhia/common/commit/b9d6daa4ec7f106cc9d96c984d09a286df1a63c1.

PR #134, opened 2018-06-17
Make explicit require statements

DzyubSpirit:
According to browser build problem https://github.com/metarhia/globalstorage/issues/106#issuecomment-391463844 `require`s should be defined explicitly because webpack can't figure out dependecies and doesn't include these modules

### DIFF
--- a/common.js
+++ b/common.js
@@ -13,7 +13,7 @@ const submodules = [
   require('./lib/events'), // Events and emitter
   require('./lib/units'), // Units conversion
   require('./lib/network'), // Network utilities
-  require('./lib/id'), // Kyes and identifiers
+  require('./lib/id'), // Keys and identifiers
   require('./lib/sort'), // Sort compare functions
   require('./lib/cache'), // Cache (enhanced Map)
   require('./lib/mp'), // Metaprogramming
--- a/common.js
+++ b/common.js
@@ -1,22 +1,22 @@
 'use strict';
 
 const submodules = [
-  'utilities', // Common utilities
-  'math', // Math common function
-  'array', // Arrays manipulations
-  'data', // Data structures manipulations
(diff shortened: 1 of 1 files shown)


aqrln:
@tshemsedinov I believe so, yes. Given that the list of modules is actually static and, roughly speaking, we just need to help webpack understand that the `path` name may only be bound to one of those values, it should be feasible to do with webpack's `require.context` and/or (maybe) `webpack.ContextReplacementPlugin`. Also, constructing the path and requiring the module would probably need to be done in one step, like this:

```diff
-].map(path => './lib/' + path).map(require);
+].map(path => require('./lib/' + path));
```

Alternatively, specifying all the modules in the webpack config as a multi-main entry might also work somehow, but I'm not sure about that.

common.js
```diff
@@ -1,22 +1,22 @@
...
+  require('./lib/data'), // Data structures manipulations
+  require('./lib/strings'), // Strings utilities
+  require('./lib/time'), // Data and Time functions
+  require('./lib/fp'), // Functional programming
+  require('./lib/oop'), // Object-oriented programming
+  require('./lib/callbacks'), // Callback utilities
+  require('./lib/events'), // Events and emitter
+  require('./lib/units'), // Units conversion
+  require('./lib/network'), // Network utilities
+  require('./lib/id'), // Kyes and identifiers
```
aqrln: While you are there (if this is eventually going to be merged), or as a TODO note for a new PR: s/Kyes/Keys

DzyubSpirit:
Nice to know that webpack supports altering require context 👍 But I think that simplicity of explicit requirements is better than beaty of code + more complex logic of require context in this case.

tshemsedinov:
Closed in favor of https://github.com/metarhia/common/pull/178

Issue #147, opened 2018-08-03
Storage identification functions


tshemsedinov:
- [x] `idToChunks(id: number): Array`
- [x] `idToPath(id: number): string`
- [ ] `pathToId(path: string): number`

tshemsedinov:
See https://github.com/metarhia/common/pull/146 @SemenchenkoVitaliy here is two function to convert identifier to storage key, you can implement third one.

PR #151, opened 2018-08-04
Implement pathToId and add tests for it

SemenchenkoVitaliy:
For global storage and file storage service

Refs: https://github.com/metarhia/common/issues/147


lib/id.js
```diff
@@ -137,6 +137,19 @@ const idToPath = id => {
...
+const pathToId = path => {
+  const chunks = path.split('/');
+  const count = chunks.length;
+  let hex = '0x';
+  let i;
+
+  for (i = count - 1; i >= 0; i--) {
+    hex += chunks[i];
+  }
+  const id = parseInt(hex);
```
tshemsedinov: I'd rather use `parseInt(hex, 16)`

lib/id.js
```diff
@@ -137,6 +137,19 @@ const idToPath = id => {
   return path;
 };
 
+const pathToId = path => {
+  const chunks = path.split('/');
+  const count = chunks.length;
+  let hex = '0x';
+  let i;
+
```
tshemsedinov: Empty line is not needed here

lib/id.js
```diff
@@ -137,6 +137,19 @@ const idToPath = id => {
...
+  for (i = count - 1; i >= 0; i--) {
+    hex += chunks[i];
+  }
+  const id = parseInt(hex);
+  return id;
```
lundibundi: Why put id to a variable and not just `return parseInt(hex, 16)`?

test/id.js
```diff
@@ -44,6 +44,14 @@ api.metatests.case('Common / id', {
     [123456789,        'cd15/075b'],
     [123456789123,     '1a83/be99/001c'],
     [9007199254740991, 'ffff/ffff/ffff/001f']
+  ],
+  'common.pathToId': [
+    ['0000/0000',           0],
+    ['0001/0000',           1],
```
lundibundi: Add a few corner cases that make sure that id will be different for '0001/0000' and '0000/0001' and such.

lib/id.js
```diff
@@ -137,6 +137,19 @@ const idToPath = id => {
   return path;
 };
 
+const pathToId = path => {
+  const chunks = path.split('/');
+  const count = chunks.length;
+  let hex = '0x';
+  let i;
+
+  for (i = count - 1; i >= 0; i--) {
```
lundibundi: I don't think it's worth to cache length as you only use it once here.

tshemsedinov:
Next time use `commit --fixup` please for such a changes so it will autosquash on landing

tshemsedinov:
One more hint: you can create branch in metarhia not separate repo fork in your account

lib/id.js
```diff
@@ -137,6 +137,16 @@ const idToPath = id => {
   return path;
 };
 
+const pathToId = path => {
+  const chunks = path.split('/');
+  let hex = '0x';
```
belochub: You don't need to add `0x` at the start of the number string if you are using `parseInt()` providing the base to it.
Also, you can make this function a one-liner that way.

tshemsedinov: Sorry, I already landed these PR. But yes, `split.reverse.join` may turn this func to one-lien. In spite of this such an implementation is good enough and may even be faster. @belochub

tshemsedinov:
Landed in f77f5dc

tshemsedinov:
@SemenchenkoVitaliy if you closes issue by commit, you should add `Closes: URL` tag, not `Refs: URL` so github will automatically close that issue.

SemenchenkoVitaliy:
Ok, understood

lib/id.js
belochub: @tshemsedinov, you've forgotten to land both commits or to squash commits before landing.

belochub: Also, you can make it a one liner that way:
```js
parseInt(path.split('/').reduceRight((hex, chunk) => hex + chunk), 16);
```

tshemsedinov: @belochub thanks, I'll add it as a separate commit

PR #154, opened 2018-08-06
Add owasp authentication library

### LINKED ISSUE #148: Implement password strength OWASP recomendations
- [x] `class AuthenticationStrength { get valid() {}, strength, hints }`
- [x] `AuthenticationStrength = checkPassword(password)`
- [x] `AuthenticationStrength = checkLogin(login)`
- [x] `AuthenticationStrength = checkLoginPassword(login, password)`

Refs: https://www.owasp.org/index.php/Authentication_Cheat_Sheet

bugagashenkj:
Close #148


lib/authentication.js
```diff
@@ -0,0 +1,147 @@
+'use strict';
+
+const MAX_PASS_LENGTH = 128;
+const MIN_PASS_LENGTH = 10;
+const MIN_PHRASE_LENGTH = 20;
+const MIN_OPTIONAL_TESTS = 3;
+
+const MIN_LOGIN_LENGTH = 3;
+const MAX_LOGIN_LENGTH = 30;
```
tshemsedinov: If user will use email for login we need at least 64 chars (gmail have 255 max email length)

lib/authentication.js
```diff
@@ -0,0 +1,147 @@
+'use strict';
+
+const MAX_PASS_LENGTH = 128;
+const MIN_PASS_LENGTH = 10;
+const MIN_PHRASE_LENGTH = 20;
+const MIN_OPTIONAL_TESTS = 3;
+
+const MIN_LOGIN_LENGTH = 3;
```
tshemsedinov: Are there recommendations to have 3 letter logins? I think we need at least 6 chars but we need to follow recommendations if any standardization authority have it?

lib/authentication.js
```diff
@@ -0,0 +1,147 @@
...
+    } else if (password.length > MAX_PASS_LENGTH) {
+      return `The password must be less than ${MAX_PASS_LENGTH} characters`;
+    }
+  },
+
+  password => {
+    if (/(.)\1{2,}/.test(password)) {
+      return 'The password can not contain a sequence of repeated chars';
+    }
+  }
```
tshemsedinov:
```
const passRequiredTests = [
  [
    password => password.length < MIN_PASS_LENGTH,
    `The password must be longer than ${MIN_PASS_LENGTH} characters`
  ], [
    password => password.length > MAX_PASS_LENGTH,
    `The password must be less than ${MAX_PASS_LENGTH} characters`
  ], [
    password => /(.)\1{2,}/.test(password),
    'The password can not contain a sequence of repeated chars'
  ]
];
```

lib/authentication.js
```diff
@@ -0,0 +1,147 @@
...
+    if (login.includes(password)) {
+      return 'The login can not contain a password';
+    } else if (password.includes(login)) {
+      return 'The password can not contain a login';
+    }
+  }
+];
+
+const createResult = hints => (
+  new (class AuthenticationStrength {
```
tshemsedinov: Do not put class to IIFE

lib/authentication.js
```diff
@@ -0,0 +1,147 @@
...
+  (login, password) => {
+    if (login.includes(password)) {
+      return 'The login can not contain a password';
+    } else if (password.includes(login)) {
+      return 'The password can not contain a login';
+    }
+  }
+];
+
+const createResult = hints => (
```
tshemsedinov: What is hints? Naming and data structure is not clear

lib/authentication.js
```diff
@@ -0,0 +1,147 @@
...
+  new (class AuthenticationStrength {
+    get hints() { return hints; }
+    get valid() {
+      return hints.login ? !hints.login.length : true &&
```
tshemsedinov: Terrible ternary enclosure, decompose it to intermediate variables

tshemsedinov: Do not use `(expression1) ? true : (expression2)`, logical expression is already boolean type, so you can simply write `expression1 && expression2`, or `if (expression1) return true; ... /* further checks */`

lib/authentication.js
```diff
@@ -0,0 +1,145 @@
...
+  ]
+};
+
+class AuthenticationStrength {
+  constructor(requiredHints, optionalHints = []) {
+    this.hints = {
+      required: requiredHints,
+      optional: optionalHints
+    };
+  }
```
tshemsedinov:
```
  constructor(required, optional = []) {
    this.hints = { required, optional };
  }
```

lib/authentication.js
```diff
@@ -0,0 +1,145 @@
...
+  get valid() {
+    return !this.hints.required.length;
+  }
+  get strength() {
+    return !this.hints.optional.length;
```
tshemsedinov: I think this should be number

tshemsedinov: I think strength will be:
- Very weak 0 - 24%
- Weak 25 - 39%
- Good 40 - 59%
- Strong 60 - 79%
- Very strong 80 - 100%

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+      return 'Very strong';
+    }
+  }
+}
+
+const makeTest = (
+  // Function that checks the arguments on a test suite
+  tests, // Array, array of password/login tests
+  testGroups, // Array, [ [[requiredTest, ...], [optionalTest, ...]], ...]
+  ...testArgm // Array, [password]/[login]/[login, password]
```
lundibundi: testArgs?

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      return 'Very strong';
+    }
+  }
+}
+
+const makeTest = (
+  // Function that checks the arguments on a test suite
+  tests, // Array, array of password/login tests
+  testGroups, // Array, [ [[requiredTest, ...], [optionalTest, ...]], ...]
```
lundibundi: I don't understand why is it possible to have multiple groups of required and optional tests? Isn't it one array of required tests and one of optional? Then it's possible to just use
```
{
  required: [],
  optional: [],
}
```

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
+'use strict';
+
+const MAX_PASS_LENGTH = 128;
+const MIN_PASS_LENGTH = 10;
+const MIN_PHRASE_LENGTH = 20;
+
+const MIN_LOGIN_LENGTH = 6;
+const MAX_LOGIN_LENGTH = 255;
+
+const passwordTests = {
```
lundibundi: These don't seem to be exported, is there a way to pass wanted tests to the appropriate functions?

lib/authentication.js
bugagashenkj: I think that the explanation by the example will be more clear. The password will be valid if (the password is longer than the minimum password length + is less than the maximum password length + there are no duplicate characters) or if (the password is greater than MIN_PHRASE_LINGTX). Each of the required conditions has its own optional conditions.
```
const testGroups = [
  [
    ['MIN_LENGTH', 'MAX_LENGTH', 'REPEAT_CHARS'],
    ['ONE_LOWERCASE_CHAR', 'ONE_NUMBER', 'ONE_UPPERERCASE_CHAR']
  ], [
    ['MIN_PHRASE_LENGTH'],
    ['REPEAT_CHARS']
  ]
];
```

lib/authentication.js
bugagashenkj: Yes. You just need to pass the names of the necessary tests to the test function.

lib/authentication.js
bugagashenkj: Yes thank you. It will be fixed.

lib/authentication.js
lundibundi: Do I understand it correctly that for 'MIN_LENGTH' optional is 'ONE_LOWERCASE_CHAR' and so on?

bugagashenkj: Yes.

lib/authentication.js
lundibundi: Maybe that's a place to use our new `Enum` class? @tshemsedinov what do you think?

lib/authentication.js
lundibundi: Could you elaborate, I proposed 2 ways.
Also, either way, I do not see why there should be such a distinction and how first and second subarrays are connected. Why can't just there be an object of 2 arrays as I suggested?

bugagashenkj: If we use an object with two arrays, it will be difficult for us to build a disjunction inside each of the arrays. Another example: if the login is an email, then it is valid and does not have optional tests, but if the login is not an email, it can have optional tests.
```
const loginTestGroups = [
  [
    ['IS_EMAIL'],
    []
  ], [
    ['MIN_LENGTH', 'MAX_LENGTH'],
    ['IS_BRUTE_FORCE_VALUE', 'REPEAT_CHARS']
  ]
];
```
How to do this with two arrays?

lundibundi: So you mean that all of the top array entries are a disjunction (it only has to pass a single of those)? What is the use case for this?

bugagashenkj: Disjunction applies only to test groups (group is array of required tests and array of optional tests). Therefore, in order for the password to be valid it must pass all the required tests, at least one group. Use case: my first example with MIN_LENGTH and MIN_PHRASE_LENGTH. https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Password_Length

lundibundi: Then I'd still suggest to use {  required: [],  optional: [] } for subarrays to avoid overcomplicating the syntax.
So it will look something like
```js
[
  {
    required: ['IS_EMAIL'],
  },
  {
    required: ['MIN_LENGTH', 'MAX_LENGTH'],
    optional: ['IS_BRUTE_FORCE_VALUE', 'REPEAT_CHARS']
  }
]
```

lib/authentication.js
tshemsedinov: agree :+1:

lib/authentication.js
bugagashenkj: Ok, thanks. I also think that it will look better.

lib/authentication.js
tshemsedinov: Please add here example with enums to compare

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+  ]
+};
+
+const loginTests = {
+  MIN_LENGTH: [
+    login => login.length < MIN_LOGIN_LENGTH,
+    `The login must be longer than ${MIN_LOGIN_LENGTH} characters`
+  ],
+  MAX_LENGTH: [
+    login => login.length > MAX_LOGIN_LENGTH,
```
lundibundi: `MAX_LOGIN_LENGTH` doesnt't seem to be configurable, shouldn't it be configurable? (same goes for `MIN_LOGIN_LENGTH`, `MIN_PASS_LENGTH`, `MAX_PASS_LENGTH`, `MIN_PHRASE_LENGTH`)

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+  ]
+};
+
+class AuthenticationStrength {
+  constructor(
+    hints, // Array, [{ required: Array, optional: Array }, ...]
+    bestPercent // Number, percent of best test group
+  ) {
+    this.hints = hints;
+    this.bestPercent = bestPercent;
```
lundibundi: Maybe `compliancePercent` or just `compliance`?

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+);
+
+const checkLoginPassword = (
+  // Function that tests the login with password
+  login, // String, login for testing
+  password, // String, password for testing
+  testsGroups // Array, [{ required: [testNames], optional: [testNames] }]
+  // Returns: AuthenticationStrength instance
+) => (
+  makeTest(loginPasswordTests, testsGroups, login, password)
```
lundibundi: What's the use of passing `loginPasswordTests`, `loginTests`, `passwordTests` everywhere as they are global in this file and always the same?

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+  login, // String, login for testing
+  testGroups // Array, [{ required: [testNames], optional: [testNames] }]
+  // Returns: AuthenticationStrength instance
+) => (
+  makeTest(loginTests, testGroups, login)
+);
+
+const checkPassword = (
+  // Function that tests the password
+  password, // String, password for testing
```
lundibundi: s/password for testing/password to test/ ? (If you agree then same in other places)

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+      return 'Very strong';
+    }
+  }
+}
+
+const makeTest = (
+  // Function that checks the arguments on a test suite
+  tests, // Array, array of password/login tests
+  testGroups, // Array, [{ required: [testNames], optional: [testNames] }]
+  ...testArgs // Array, [password]/[login]/[login, password]
```
lundibundi: Could you please reorder it to
`[login] / [password] / [login, password]`
?
=)

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+    bestPercent // Number, percent of best test group
+  ) {
+    this.hints = hints;
+    this.bestPercent = bestPercent;
+  }
+  get valid() {
+    return !!this.bestPercent;
+  }
+  get strength() {
+    if (this.bestPercent < 25) {
```
lundibundi: I'd say to drop this to `< 20` and then use an Array of strength and 
`strengths[Math.min(Math.floor(this.bestPercent / 20), 4)]` or something.
@tshemsedinov what do you think?

lib/authentication.js
```diff
@@ -0,0 +1,172 @@
...
+    'The password must contain uppercase char'
+  ],
+  ONE_NUMBER: [
+    password => !/[0-9]/.test(password),
+    'The password must contain number'
+  ],
+  ONE_SPECIAL_CHAR: [
+    password => !/[^A-Za-z0-9]/.test(password),
+    'The password must contain special char'
+  ]
```
lundibundi: All of the `ONE` tests should probably be generalized to `N` chars, wdyt?

lib/authentication.js
bugagashenkj: All tests and their parameters are unconfigurable.

lib/authentication.js
bugagashenkj: You can find out if your data passes the selected tests or not. I think I did not quite understand this question.

lib/authentication.js
bugagashenkj: Firstly, our tests do not have input parameters and I do not quite understand how you want to set the n. Secondly, the specifications specify tests for one character.

bugagashenkj:
@lundibundi, please look at the new implementation of tests related to Unicode categories.

lib/authentication.js
lundibundi: Nevermind, I misunderstood the `makeTest`.

lib/authentication.js
```diff
@@ -0,0 +1,220 @@
...
+
+const EMAIL_REGEXP = new RegExp([
+  "^[-a-z0-9!#$%&'*+/=?^_`{|}~]+(?:\.[-a-z0-9!#$%&'*+/=?^_`{|}~]+)*@(?:[a-z0-", // eslint-disable-line
+  "9]([-a-z0-9]{0,61}[a-z0-9])?\.)*(?:aero|arpa|asia|biz|cat|com|coop|edu|gov", // eslint-disable-line
+  "|info|int|jobs|mil|mobi|museum|name|net|org|pro|tel|travel|[a-z][a-z])$" // eslint-disable-line
+].join(''));
+
+const passwordTests = {
+  MIN_LENGTH: [
+    password => password.length < MIN_PASS_LENGTH,
```
lundibundi: If I understand correctly all tests check for 'falseness', therefore tests passes if it returns 'false'? Could you please reverse this as it is confusing, you expect something to pass if it returns true, not false.

lib/authentication.js
```diff
@@ -0,0 +1,220 @@
...
+    password => password.length < MIN_PHRASE_LENGTH,
+    `The passphrase must be longer than ${MIN_PHRASE_LENGTH} characters`
+  ],
+  REPEAT_CHARS: [
+    password => /(.)\1{2,}/.test(password),
+    'The password can not contain a sequence of repeated chars'
+  ],
+  ONE_LOWERCASE_CHAR: [
+    password => !LOWERCASE_REGEXP.test(password),
+    'The password must contain lowercase char'
```
lundibundi: 'a lowercase char' or 'at least one lowercase char'? here and below.

lib/authentication.js
```diff
@@ -0,0 +1,220 @@
...
+    password => !UPPERCASE_REGEXP.test(password),
+    'The password must contain uppercase char'
+  ],
+  ONE_NUMBER: [
+    password => !/[0-9]/.test(password),
+    'The password must contain number'
+  ],
+  ONE_SPECIAL_CHAR: [
+    password => !/[^A-Za-z0-9]/.test(password),
+    'The password must contain special char'
```
lundibundi: This message doesn't tell the caller what's expected (what chars are 'special'). Also, I don't think that you should test special chars as everything not in those groups because now, even russian letter 'э' or anything is a special char.

lib/authentication.js
```diff
@@ -0,0 +1,220 @@
...
+  ],
+  REPEAT_CHARS: [
+    password => /(.)\1{2,}/.test(password),
+    'The password can not contain a sequence of repeated chars'
+  ],
+  ONE_LOWERCASE_CHAR: [
+    password => !LOWERCASE_REGEXP.test(password),
+    'The password must contain lowercase char'
+  ],
+  ONE_UPPERERCASE_CHAR: [
```
lundibundi: s/ONE_UPPERERCASE_CHAR/ONE_UPPERCASE_CHAR/

lib/authentication.js
```diff
@@ -0,0 +1,220 @@
...
+      const optTestsCount = optional.length;
+      const optHintsCount = optionalHints.length;
+      const percent = 100 * (optTestsCount - optHintsCount) / optTestsCount;
+      compliance = Math.max(compliance, percent);
+    }
+
+    hints.push({ required: requiredHints, optional: optionalHints });
+  });
+
+  hints.sort((prevHints, hints) => {
```
lundibundi: What's this sorting for? The only way for the user to establish a connection between test and hint is by the order of them in the resulting array and this breaks it. Maybe we shouldn't do that? Also, see the general review result I'll add on this there.

bugagashenkj: Sorting was performed to sort the test results by the number of required subtests for the test. But I agree that it is better to leave a link between the input and output data.

bugagashenkj:
@lundibundi About the number of characters in regular expressions. 2063 is the number of characters. In the regular expression there are ranges of symbols "a-z".

bugagashenkj:
@tshemsedinov I have a question about password verification for the standard password. Do we need such a test? If so, how do you want to keep a list of standard passwords? In the source file?

bugagashenkj:
1. @lundibundi if I understand correctly, then you want to create sets of ranges of char codes that fall into the category? This will lead to an increase in the file size and will complicate the support of the set, since with any changes in the unicode database it will be necessary to rebuild the set. 
3. @lundibundi Before, @tshemsedinov  did not want to make the tests configurable. If we make them configurable, then there is no guarantee that the password or login that passed test is really reliable. Do we trust the person who configures the test? Why create variations of tests that are not in the spicifications? If all the same, we will make the tests configurable, how do @tshemsedinov want to pass arguments to each test?

lundibundi:
1. @bugagashenkj Something like that, as I said, please, contact @belochub for further guidance, he has implemented such case before. No it won't be harger to support than what you have, because there will just be a script that generates those ranges directly from the Unicode website and you will only have to run that script to update. If you use regexps and such you will have to manually update the list every time. Also, regexp check will be much (I'd say at least a few times, if not tenths of times slower) than range checks.

2. @tshemsedinov, please clarify for @bugagashenkj the idea. Also, as I said before, we will have default checks that are there if a user didn't provide any configuration. And if he did provide it then we assume that he is competent enough to do that.

bugagashenkj:
1. @lundibundi see the new implementation. Where it is better to save the parser?
3. It is necessary to understand at what level it is necessary to set the parameters for the tests. The user can customize the settings for all of his test groups. In this case, it will be convenient for us to export not 3 functions, but a class with mutable parameters. If the user wants to configure test parameters for each group separately, then I think it will be convenient to make 
```
{
  required: [...],
  optional: [...],
  options: {
    MIN_PASSWORD_LENGTH: 8,
    MIN_LOWERCASE_CHARS: 2
  }
}
If the user wants to configure each test, you can pass not the name of the test, but an array with the name and parameters.
```
{
  required: [ [MIN_LENGTH, 2], [MAX_LENGTH, 2] },
  optional: [...],
}
```
What level of configuration tests are we interested in?
Do you think there is a need to take test data (character sets, standard passwords) in separate files?

File: lib/authentication.js [hunk 24]
```diff
@@ -0,0 +1,331 @@
...
+
+const MAX_PASS_LENGTH = 128;
+const MIN_PASS_LENGTH = 10;
+const MIN_PHRASE_LENGTH = 20;
+
+const MIN_LOGIN_LENGTH = 6;
+const MAX_LOGIN_LENGTH = 255;
+
+const LOWERCASE_CHAR_CODES = {
+  charCodeRanges: [
```
lundibundi: How did you generate those tables? It's better to have a script to do this, as for each new Unicode version this will have to change.

lib/authentication.js
```diff
@@ -0,0 +1,331 @@
...
+    42802, 42804, 42806, 42808, 42810, 42812, 42814, 42816, 42818, 42820,
+    42822, 42824, 42826, 42828, 42830, 42832, 42834, 42836, 42838, 42840,
+    42842, 42844, 42846, 42848, 42850, 42852, 42854, 42856, 42858, 42860,
+    42862, 42873, 42875, 42880, 42882, 42884, 42886, 42891, 42893, 42896,
+    42898, 42902, 42904, 42906, 42908, 42910, 42912, 42914, 42916, 42918,
+    42920, 42934, 42936, 120778
+  ]
+};
+
+const EMAIL_REGEXP = new RegExp(
```
lundibundi: Could you tell, where did you get this regex and what are the reasons for limiting possible domains here?
I'd say to use something like
```
\A[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@
(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\z
```
Refs: https://www.regular-expressions.info/email.html (regex copy-pasted from there) and https://tools.ietf.org/html/rfc5322#section-3.4

lib/authentication.js
```diff
@@ -0,0 +1,331 @@
...
+  ],
+  REPEAT_CHARS: [
+    password => !/(.)\1{2,}/.test(password),
+    'The password can not contain a sequence of repeated chars'
+  ],
+  ONE_LOWERCASE_CHAR: [
+    password => {
+      for (let position = 0; position < password.length; position++) {
+        const charCode = password.charCodeAt(position);
+        for (const [min, max] of LOWERCASE_CHAR_CODES.charCodeRanges) {
```
lundibundi: Could you use binary search for such searches?

lundibundi:
Added a few comments.

1. Character ranges are better stored in a separate file(s) as this will make them much easier to update/generate. 
> Where it is better to save the parser?

What do you mean by 'parser'? If you mean Unicode tables parser to generate ranges, then add a root `tools` folder and add them there (so it will be `common-folder/tools`). See https://github.com/metarhia/mdsf/blob/master/tools/make-unicode-tables.js for reference.

2. I don't like the separation of options from tests in the first case and it's hard to understand what you are confguring if you just use an array for everything (second case).
I'd suggest having each test specified as
`'Name'` or 
```
{
  name: 'Name',
  param1: 13,
  param2: 42,
}
```
Therefore if you have just string you assume all default parameters for this test (where string is a name), if there is an object you merge defaults with the ones in this object and the name of the test will be denoted as object.name.

So that test function will accept:
```
{
  required: [ test1, test2 ],
  optional: [ test3 ],
}
```
To give an example:
```
{
  required: [ 'MIN_LENGTH', { name: 'MAX_LENGTH', length: 64 }, { name: 'NUMBERS', count: 3 } ],
  optional: [ { name: 'SPECIAL_CHARS', count: 1} ],
}
```

Here I replaced ONE_SPECIAL_CHAR and ONE_NUMBER with generic SPECIAL_CHARS and NUMBERS accordingly.


------------------

Also, @tshemsedinov ping, whyt of

I think we need to return es6 Map (or at least object) of testname-to-hints (i.e. `{ 'MAX_LENGTH': 'Password exceeds maximum allowed length' }`) from the test functions as this will allow a user to simply understand which test produced which error and still get all of the hints via simple `Object.values` if needed. (now we have just an array of hints returned)

lib/authentication.js
lundibundi: The simple check in the link
```
     Check for presence of at least one @ symbol in the address
    Ensure the local-part is no longer than 64 octets
    Ensure the domain is no longer than 255 octets
    Ensure the address is deliverable
```
seem to allow for much more 'invalid'/non-existent emails than is allowed by the spec tools.ietf.org/html/rfc5322#section-3.4 and tools.ietf.org/html/rfc5321
Though this requires additional checking.

belochub: @lundibundi, this validation practice follows the RFC 5321, and I don't see how it allows invalid or non-existent emails, can you elaborate, please?

bugagashenkj:
@lundibundi where do you think it's best to store files with unicode categories?

lundibundi:
I think `lib/utils/unicode-tables.js` might be fine though I'm not sure, @belochub wdyt?
Also, tables should be generated at install-time. If you know how to do it (you can also check out metarhia-mdsf for the example) you could add it in this PR, though it is not required and can be implemented later in the follow-up PRs.

bugagashenkj:
@lundibundi please see the new configurable tests.

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+      return !regexp.test(password);
+    },
+    options => 'The password can not contain a sequence of more than ' +
+      `${options.number} repeat chars`,
+    { number: 2 }
+  ],
+  LOWERCASE_CHARS: [
+    (password, option) => {
+      let lowercaseChars = 0;
+      for (let position = 0; position < password.length; position++) {
```
lundibundi: Note: If [this](https://github.com/metarhia/common/pull/187) gets landed you could replace this.

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+        }
+      }
+    },
+    options => `The password must contain more than ${options.number} ` +
+      'uppercase chars',
+    { number: 1 }
+  ],
+  NUMBERS: [
+    (password, options) => {
+      const NUMBERS = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10'];
```
lundibundi: I don't think '10' is correct here (at least your check below won't ever check for '10'), is it?

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+          if (++numbers === options.number) return true;
+        }
+      }
+    },
+    options => `The password must contain more than ${options.number} numbers`,
+    { number: 1 }
+  ],
+  SPECIAL_CHARS: [
+    (password, options) => {
+      const SPECIAL_CHARS = [' ', '!', '"', '#', '$', '%', '&', '\'', '(', ')',
```
lundibundi: Also, maybe use strings here and above?

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+    login => {
+      const EMAIL_REGEXP = new RegExp(
+        '^[a-z0-9!#$%&\'*+/=?^_`{|}~-]+(?:.[a-z0-9!#$%&\'*+/=?^_`{|}~-]+)*@(' +
+        '?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$'
+      );
+      const MAX_EMAIL_LENGTH = 255;
+      const MAX_LOCAL_PART_LENGTH = 64;
+
+      const [localPart, ] = login.split('@');
+      return EMAIL_REGEXP.test(login) && login.length <= MAX_EMAIL_LENGTH &&
```
lundibundi: Nit: could you make it
```js
      return EMAIL_REGEXP.test(login) &&
        login.length <= MAX_EMAIL_LENGTH &&
        localPart.length <= MAX_LOCAL_PART_LENGTH;
```
?

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+    } else {
+      return 'Very strong';
+    }
+  }
+}
+
+const makeTest = (
+  // Function that checks the arguments on a test suite
+  tests, // Array, array of password/login tests
+  testGroups, // Array, [{ required: [testNames], optional: [testNames] }]
```
lundibundi: I think this arg-note needs correction (and in other places).

lundibundi:
Also, I only skimmed through Unicore-ranges generation, I hope @belochub could take a closer look.

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+
+  testGroups.forEach(({ required, optional = [] }) => {
+    const requiredHints = test(required);
+    const optionalHints = test(optional);
+
+    if (!requiredHints.length) {
+      const optTestsCount = optional.length;
+      const optHintsCount = optionalHints.length;
+      const percent = 100 * (optTestsCount - optHintsCount) / optTestsCount;
+      compliance = Math.max(compliance, percent);
```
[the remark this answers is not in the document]
belochub: On the second thought, can you please explain, why exactly do you set a total `compliance` value to the maximum percent value from all of the optional tests and not to some mean value or a minimum percent value?

lib/utils/unicode-categories.js
```diff
@@ -0,0 +1,48 @@
...
...[11491,11492],11500,11502,11507,[11520,11565],42561,42563,42565,42567,42569,42571,42573,42575,42577,42579,42581,42583,42585,42587,42589,42591,42593,42595,42597,42599,42601,42603,42605,42625,42627,42629,42631,42633,42635,42637,42639,42641,42643,42645,42647,42649,42651,42787,42789,42791,42793,42795,42797,[42799,42801],42803,42805,42807,42809,42811,42813,42815,42817,42819,42821,42823,42825,42827,42829,42831,42833,42835,42837,42839,42841,42843,42845,42847,42849,42851,42853,42855,42857,42859,42861,42863,[42865,42872],42874,42876,42879,42881,42883,42885,42887,42892,42894,42897,[42899,42901],42903,42905,42907,42909,42911,42913,42915,42917,42919,42921,42927,42933,42935,42937,43002,[43824,43866],[43872,43967],[64256,64279],[65345,65370],[66600,66639],[66776,66811],[68800,68850],[71872,71903],[93792,93823],[119834,119859],[119886,119911],[119938,119963],[119990,120015],[120042,120067],[120094,120119],[120146,120171],[120198,120223],[120250,120275],[120302,120327],[120354,120379],[120406,120431],[120458,120485],[120514,120538],[120540,120545],[120572,120596],[120598,120603],[120630,120654],[120656,120661],[120688,120712],[120714,120719],[120746,120770],[120772,120777],120779,[125218,125251]]}
```
belochub: This file is missing a newline at the end.

tools/unicode-category-parser.js
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+
+  linereader.on('close', () => {
+    resultObject.addRange(range);
+    const resultData = getFileHeader() + JSON.stringify(resultObject);
+    fs.writeFileSync(OUTPUT_PATH, resultData);
+  });
+});
+
```
belochub: Redundant empty line.

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+) => (
+  makeTest(loginPasswordTests, testsGroups, login, password)
+);
+
+module.exports = {
+  checkLogin,
+  checkPassword,
+  checkLoginPassword
+};
+
```
belochub: Redundant empty line.

bugagashenkj:
@lundibundi please see the code after the fixup.

test/authentication.js
```diff
@@ -0,0 +1,336 @@
...
+  const validPasswords = [
+    'password1234',
+    'passphrase1234567890'
+  ];
+  validPasswords.forEach(password => {
+    const passedResult = common.checkPassword(password);
+    test.strictSame(passedResult.valid, true);
+  });
+
+  const fakePasswords = [
```
lundibundi: s/fakePasswords/invalidPasswords

And same later.

tools/unicode-category-parser.js
```diff
@@ -1,3 +1,25 @@
+//MIT License
```
lundibundi: Should have spaces between // and text.

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+const checkLogin = (
+  // Function that tests the login
+  login, // String, login to test
+  testGroups = [{
+    required: ['MIN_LENGTH', 'MAX_LENGTH']
+  }, {
+    required: ['IS_EMAIL']
+  }]
+  // Returns: AuthenticationStrength instance
```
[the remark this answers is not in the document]
bugagashenkj: I think these functions return the result of makeTest function.

.eslintignore
```diff
@@ -1,2 +1,3 @@
 node_modules/
 coverage/
+dist
```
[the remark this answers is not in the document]
bugagashenkj: Now this is already in master and after the rebase is not a change.

lib/authentication.js
```diff
@@ -0,0 +1,255 @@
...
+
+class AuthenticationStrength {
+  constructor(
+    hints, // Array, [{ required: Array, optional: Array }, ...]
+    compliance // Number, percent of best test group
+  ) {
+    this.hints = hints;
+    this.compliance = compliance;
+  }
+  get valid() {
```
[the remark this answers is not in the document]
bugagashenkj: Valid is a property that indicates the passed password required tests or not. Strength indicate the level of complexity only of the valid password. It is an additional property that depends on the valid.

lib/authentication.js
```diff
@@ -0,0 +1,247 @@
...
+    } else {
+      return true;
+    }
+  }
+  return false;
+};
+
+const stringIncludesChars = (str, ranges, charsNumber) => {
+  let number = 0;
+  for (let index = 0; index <= str.length - charsNumber + number; index++) {
```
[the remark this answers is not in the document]
lundibundi: I think it'd be better to at least extract `str.length - charsNumber` to a variable.

lib/authentication.js
```diff
@@ -0,0 +1,247 @@
...
+        compliance < 0.4 ? 'Weak' :
+          compliance < 0.6 ? 'Good' :
+            compliance < 0.8 ? 'Strong' : 'Very strong';
+  }
+}
+
+const makeTest = (
+  // Function that checks the arguments on a test suite
+  tests, // Array, array of password/login tests
+  testGroups, // Array, [{ required: [], optional: [] }]
```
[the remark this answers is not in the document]
lundibundi: That's what I've been concerned about. I assume this is due to the way OWASP provides their recommendations (password is 'minLen+symbols+...' OR 'isPhrase') but IMO it is better to make user manually check that if they need to support multiple 'acceptable' conditions (this would be as simple as calling this function twice with different testGroup) instead of introducing this complexity to the implementation.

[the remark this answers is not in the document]
belochub: @lundibundi I can't seem to find the recommendation you mentioned in the article you provided, can you quote it, please?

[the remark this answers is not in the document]
lundibundi: > Longer passwords provide a greater combination of characters and consequently make it more difficult for an attacker to guess.
Minimum length of the passwords should be enforced by the application.
Passwords shorter than 10 characters are considered to be weak (NIST SP800-132).
While minimum length enforcement may cause problems with memorizing passwords among some users, applications should encourage them to set passphrases (sentences or combination of words) that can be much longer than typical passwords and yet much easier to remember.
Maximum password length should not be set too low, as it will prevent users from creating passphrases. Typical maximum length is 128 characters.
Passphrases shorter than 20 characters are usually considered weak if they only consist of lower case Latin characters.

While not an explicit reccomendation this assumes that if user sets a password (not a passphrase) then minLength should be at least 10 while if he uses a passphrase it should be at least 20.
I think @bugagashenkj can clarify better the reasons behind such implementation.

Though as I mentioned I'd prefer to not allow multiple testGroups and instead make user do it manually if they need it.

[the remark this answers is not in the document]
bugagashenkj: From the recommendations of owasp:
1. As already said the password is valid, if it is longer than 10 characters, it does not have duplicate characters... OR if it is longer of 20 characters(passphrase).
2. Login is valid if it is longer than minLength... OR if it is email.
But beyond that, it can be any custom test configuration.

lib/authentication.js
```diff
@@ -0,0 +1,247 @@
...
+    const optionalHints = test(optional);
+
+    if (!requiredHints.length && compliance !== 1) {
+      valid = true;
+      let ratio = 1;
+      if (optional.length) ratio -= optionalHints.length / optional.length;
+      if (compliance < ratio) compliance = ratio;
+    }
+
+    hints.push({ required: requiredHints, optional: optionalHints });
```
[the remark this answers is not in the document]
bugagashenkj: The user can have with a valid password, which will be very weak (will not pass the optional tests). And the user can try changing the password using another test group.
Example: the user have a password longer than minLength, and shorter than MaxLenght -> the password is valid, but user did not use: uppercase chars, numbers, special chars -> the password is valid, but weak. How can a user make a password strong? He can add something from (uppercase, number, special char), or make the password longer than passphrase minLenght (maybe it's easier for him). For this he must to see passphrase required hint.

lib/authentication.js
```diff
@@ -0,0 +1,247 @@
...
+      return true;
+    }
+  }
+  return false;
+};
+
+const stringIncludesChars = (str, ranges, charsNumber) => {
+  let number = 0;
+  for (let index = 0; index <= str.length - charsNumber + number; index++) {
+    const charCode = str.codePointAt(index);
```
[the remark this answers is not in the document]
bugagashenkj: Sorry i don't understand last comment. @lundibundi or @belochub can you explain, please?

lib/authentication.js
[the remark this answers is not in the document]
belochub: For example, take a situation where there are no optional tests, only required tests in multiple test groups provided.
If the user provides the password that is already valid, what is the point of showing him the hints to make the password valid? He can't make it more valid or less valid, it's either valid or not.

lib/authentication.js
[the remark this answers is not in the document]
belochub: One Unicode code point can be encoded as either one or two UTF-16 code units, due to UTF-16 code units being 16 bit long.

common.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const submodules = [
+  'authentication',
```
lundibundi: `auth` maybe?

bugagashenkj:
@lundibundi please see code after refactoring.

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+      if (++number === charsNumber) return true;
+    }
+  }
+  return false;
+};
+
+const passwordTests = {
+  MIN_LENGTH: {
+    test: (password, options) => password.length >= options.minLength,
+    hint: options => 'The password must be longer than or equal to ' +
```
belochub: I think wording it like `'password must be at least N chars long'` will be better.
Also, AFAIU, these hints will also be shown to the end-user; therefore we have to somehow integrate i18n for the hints to allow them to be localized when needed.

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+      'uppercase chars',
+    options: { number: 1 }
+  },
+  MIN_NUMBERS: {
+    test: (password, options) => {
+      const NUMBERS_RANGE = [[49, 57]];
+      return stringIncludesChars(password, NUMBERS_RANGE, options.number);
+    },
+    hint: options => `The password must contain at least ${options.number} ` +
+      'numbers',
```
belochub: If the `number` is equal to 1, this will say `The password must contain at least 1 numbers`, which is incorrect. There are similar issues with all of the hints.
These issues would also be solved by enabling i18n since correct pluralization is usually a part of it.

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+const checkPassword = (
+  // Function that tests the password
+  password, // String, password to test
+  required, // Array, required tests configs
+  optional = [] // Array, optional tests configs
+  // Returns: AuthenticationStrength instance
+) => {
+  if (!required) {
+    required = ['MIN_LENGTH', 'MAX_LENGTH'];
+    optional = [
```
belochub: Why do you reassign optional tests if the user doesn't provide the required tests?
What if the user wants to provide the optional tests and use the default required tests?

lib/utils/unicode-categories.js
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2018 mdsf project authors. Use of this source code is
```
belochub: You should also mark this file as autogenerated for GitHub, to do this, create a `.gitattributes` file in the root of the repository and add the next line to it:
```git
/lib/utils/unicode-categories.js linguist-generated=true
```

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+      const MAX_LOCAL_PART_LENGTH = 64;
+
+      if (login.includes('@')) {
+        const [localPart, domain] = login.split('@');
+        return domain.length <= MAX_DOMAIN_LENGTH &&
+          localPart.length <= MAX_LOCAL_PART_LENGTH &&
+          EMAIL_REGEXP.test(login);
+      }
+    },
+    hint: () => 'Login must be email address'
```
belochub: There is an article missing before 'email', must probably be 'an email'.

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+  }
+};
+
+const loginPasswordTests = {
+  LOGIN_INCLUDES_PASSWORD: {
+    test: (login, password) => !login.includes(password),
+    hint: () => 'The login can not contain a password'
```
belochub: I don't think that you mean 'a password' here (meaning it can be any password), it should probably say 'the password' since you are referring to the password that was provided to you by the user.

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+  },
+  PASSWORD_INCLUDES_LOGIN: {
+    test: (login, password) => !password.includes(login),
+    hint: () => 'The password can not contain a login'
```
belochub: Same as above, it should probably say 'the login' here.

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+  }
+};
+
+class AuthenticationStrength {
+  constructor(
+    valid, // Boolean
+    hints, // Object, { required: Array, optional: Array }
+    compliance // Number, percent of best test group
```
belochub: Article is missing before 'best'.

common.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const submodules = [
+  'auth', // Login/password validation
```
belochub: I think it is possible to use the functionality implemented here for validation of something other than logins and passwords that is also authentication/authorization related.

lib/auth.js
```diff
@@ -0,0 +1,248 @@
...
+  }
+  return false;
+};
+
+const stringIncludesChars = (str, ranges, charsNumber) => {
+  let number = 0;
+  const utf16singleUnit = 1 << 16;
+  for (let index = 0; index <= str.length - (charsNumber - number); index++) {
+    const codePoint = str.codePointAt(index);
+    if (codePoint > utf16singleUnit) index++;
```
belochub: Must be `>=` here.

lib/auth.js
bugagashenkj: Now there are two options: the user uses the default configuration and passes only the test data to the function, or the user uses its configuration.
Now the user can not pass optional tests without passing the required ones, so if the user wants to use any of his own configurations, he must pass full test configuration.

bugagashenkj:
@belochub please review after fixup.

bugagashenkj:
@nechaido please rewiew after fixup.

test/auth.js
```diff
@@ -0,0 +1,260 @@
...
+  test.strictSame(failedResult.valid, false);
+  test.end();
+});
+
+metatests.test('test checkLogin / MIN_LENGTH', (test) => {
+  const login = 'login';
+  const passedResult = common.checkLogin(login, [
+    { name: 'MIN_LENGTH', minLength: login.length }
+  ]);
+  const failedResult = common.checkPassword(login, [
```
belochub: Must be `checkLogin` here since this is a test for this function.

lib/auth.js
```diff
@@ -0,0 +1,241 @@
...
+};
+
+const stringIncludesChars = (str, ranges, charsNumber) => {
+  let number = 0;
+  const utf16singleUnit = 1 << 16;
+  for (let index = 0; index <= str.length - (charsNumber - number); index++) {
+    const codePoint = str.codePointAt(index);
+    if (codePoint >= utf16singleUnit) index++;
+    if (unicodeRangesIncludes(ranges, codePoint)) {
+      if (++number === charsNumber) return true;
```
nechaido: You can merge this with the above line.

nechaido: Although this may be left as it is, for the sake of readability.

lib/auth.js
```diff
@@ -0,0 +1,241 @@
...
+
+class AuthenticationStrength {
+  constructor(
+    valid, // Boolean
+    hints, // Object, { required: Array, optional: Array }
+    compliance // Number, ratio of passed optional tests to all optional tests
+  ) {
+    this.valid = valid;
+    this.hints = hints;
+    this.strength = !valid ? 'Not valid' :
```
nechaido: I'd prefer to use `if() {} else if() {} ...` here.

bugagashenkj:
@lundibundi, i added braces, please approve.

PR #159, opened 2018-08-07
B tree

machendos:
This is the basic implementation of B-Tree. Implemented methods: ``add, getEqual, getLess, getLarger, getBetween``. Method ``delete`` wasnt yet implemented , but this is a matter of time. When it becomes clear in what form data will come, it will be possible to implement the building of a tree on a data set. There are questions about opimization:

- At the moment, the elements in the nodes are stored as an array. Perhaps it would be better to remake this to a one-linked list? Since we never need a quick access to an item by index. And when you search for an element, you still look at the part of the array that precedes the desired one. However, adding elements to the middle will be faster and more enjoyable
- For avoid mixins in the prototype of a tree, it is probably best to first assign all methods to some object, and then transfer them to the prototype by ``Object.assign()``?

### DIFF
--- a/lib/btree.js
+++ b/lib/btree.js
@@ -116,11 +116,12 @@ const growChild = (parent, childIndex) => {
   return joinNodes(parent, childIndex - 1, childIndex);
 };
 
-const extractLimit = (
-  node,
-  minDegree,
-  upper // Boolean: true - upperLimit, false - lowerLimit
-) => {
+// Get the minimal upper, or the maximum lover node for some given node
+//   node - some given node
+//   minDegree - value of bTree.minDegree for tree
+//   upper - boolean: true - upperLimit, false - lowerLimit
+// Result: upperLimit or lowerLimit node
+const extractLimit = (node, minDegree, upper) => {
   let currNode = node;
   while (!isLeaf(currNode)) {
     const index = upper ? 0 : currNode.length - 1;
--- a/lib/btree.js
+++ b/lib/btree.js
@@ -32,21 +32,18 @@ const binarySearch = (node, key) => {
   let end = node.length - 1;
   while (start <= end) {
     const i = (start + end) >> 1;
-    const item = node[i];
-    const itemValue = item.key;
-    if (key > itemValue) start = i + 1;
-    // the biggest element in the node has key === undefined, therefore,
-    // in order to get true for it here using reverse comparison.
-    else if (!(key >= itemValue)) end = i - 1;
+    const itemKey = node[i].key;
+    if (key > itemKey) start = i + 1;
+    else if (itemKey === undefined) end = i - 1;
+    else if (key < itemKey) end = i - 1;
     else return [true, i];
   }
   return [false, start];
 };
 
 function* inorderTraversal(start, finish, currNode) {
-  const binarySearchResult = binarySearch(currNode, start);
-  let startIndex = binarySearchResult[1];
-  if (binarySearchResult[0]) startIndex++;
+  const startIndex = start === undefined ?
+    0 : binarySearch(currNode, start)[1];
   const finishIndex = finish === undefined ?
     currNode.length - 1 :
     binarySearch(currNode, finish)[1];
--- a/lib/btree.js
+++ b/lib/btree.js
@@ -34,8 +34,7 @@ const binarySearch = (node, key) => {
     const i = (start + end) >> 1;
     const itemKey = node[i].key;
     if (key > itemKey) start = i + 1;
-    else if (itemKey === undefined) end = i - 1;
-    else if (key < itemKey) end = i - 1;
+    else if (itemKey === undefined || key < itemKey) end = i - 1;
     else return [true, i];
   }
   return [false, start];
--- a/lib/btree.js
+++ b/lib/btree.js
@@ -1,0 +1,247 @@
+'use strict';
+
+const DEFAULT_DEGREE = 6; // min degree of b-tree.
+// All vertices except the root have [degree ... 2 * degree] child nodes
+// And [degree - 1 ... 2 * degree - 1] + 1 "empty" elements
+
+class Element {
+  constructor(key, data, child = null) {
+    this.key = key;
+    this.data = data;
+    this.child = child;
+  }
+}
+
+const empty = child => new Element(undefined, undefined, child);
+
+const splitNode = (parent, index) => {
+  const node = parent[index].child;
+  const len = node.length;
+  const newLeftNode = node.splice(0, len / 2 - 1); // First half
+  const mid = node.shift();
+  newLeftNode.push(empty(mid.child));
+  mid.child = newLeftNode;
+  parent.splice(index, 0, mid);
+  return parent;
+};
+
+const isLeaf = node => !node[0].child;
+
+const binarySearch = (node, key) => {
+  let start = 0;
+  let end = node.length - 1;
+  while (start <= end) {
+    const i = (start + end) >> 1;
+    const item = node[i];
+    const itemValue = item.key;
+    if (key > itemValue) start = i + 1;
+    // the biggest element in the node has key === undefined, therefore,
+    // in order to get true for it here using reverse comparison.
+    else if (!(key >= itemValue)) end = i - 1;
+    else return [true, i];
+  }
+  return [false, start];
+};
+
+function* inorderTraversal(start, finish, currNode) {
+  const binarySearchResult = binarySearch(currNode, start);
+  let startIndex = binarySearchResult[1];
+  if (binarySearchResult[0]) startIndex++;
+  const finishIndex = finish === undefined ?
+    currNode.length - 1 :
+    binarySearch(currNode, finish)[1];
+  if (isLeaf(currNode)) {
+    for (let currIndex = startIndex; currIndex < finishIndex; currIndex++) {
+      yield currNode[currIndex].data;
+    }
+  } else {
+    for (let currIndex = startIndex; currIndex < finishIndex; currIndex++) {
+      const currElement = currNode[currIndex];
+      yield* inorderTraversal(start, finish, currElement.child);
+      yield currElement.data;
+    }
+    yield* inorderTraversal(start, finish, currNode[finishIndex].child);
+  }
+}
+
+const joinNodes = (parent, firstNodeIndex, secondNodeIndex) => {
+  const firstNode = parent[firstNodeIndex].child;
+  const secondNode = parent[secondNodeIndex].child;
+  const mid = parent.splice(firstNodeIndex, 1)[0];
+  mid.child = firstNode.pop().child;
+  secondNode.splice(0, 0, ...firstNode, mid);
+  return secondNode;
+};
+
+const growChild = (parent, childIndex) => {
+  const parentElement = parent[childIndex];
+  const node = parentElement.child;
+  const minDegree = node.length;
+  if (childIndex > 0) { // If we have left neighbor
+    const leftNeighbor = parent[childIndex - 1].child;
+    if (leftNeighbor.length > minDegree) {
+      const extractedElement =
+        leftNeighbor.splice(leftNeighbor.length - 2, 1)[0];
+      const insertedElement = new Element(
+        parent[childIndex - 1].key,
+        parent[childIndex - 1].data,
+        leftNeighbor[leftNeighbor.length - 1].child
+      );
+      leftNeighbor[leftNeighbor.length - 1].child = extractedElement.child;
+      extractedElement.child = leftNeighbor;
+      parent.splice(childIndex - 1, 1, extractedElement);
+      node.unshift(insertedElement);
+      return node;
+    }
+  }
+  if (childIndex < parent.length - 1) {
+    // If we have right neighbor
+    // parent.length - 1 means that we now on the rightmost element
+    const rightNeighbor = parent[childIndex + 1].child;
+    if (rightNeighbor.length > minDegree) {
+      const extractedElement = rightNeighbor.shift();
+      const insertedElement = new Element(
+        parentElement.key,
+        parentElement.data,
+        node[node.length - 1].child
+      );
+      node[node.length - 1].child = extractedElement.child;
+      extractedElement.child = node;
+      parent.splice(childIndex, 1, extractedElement);
+      node.splice(node.length - 1, 0, insertedElement);
+      return node;
+    }
+    return joinNodes(parent, childIndex, childIndex + 1);
+  }
+  return joinNodes(parent, childIndex - 1, childIndex);
+};
+
+const extractLimit = (
+  node,
+  minDegree,
+  upper // Boolean: true - upperLimit, false - lowerLimit
+) => {
+  let currNode = node;
+  while (!isLeaf(currNode)) {
+    const index = upper ? 0 : currNode.length - 1;
+    let nextNode = currNode[index].child;
+    if (nextNode.length === minDegree) {
+      nextNode = growChild(currNode, index);
+    }
+    currNode = nextNode;
+  }
+  const index = upper ? 0 : currNode.length - 2;
+  return currNode.splice(index, 1)[0];
+};
+
+const deleteElement = (node, elementIndex, minDegree) => {
+  const element = node[elementIndex];
+  const deletedData = element.data;
+  const leftChild = element.child;
+  const rightChild = node[elementIndex + 1].child;
+  if (isLeaf(node)) {
+    return node.splice(elementIndex, 1)[0].data;
+  }
+  if (leftChild.length > minDegree) {
+    const lowerLimit = extractLimit(leftChild, minDegree, false);
+    element.key = lowerLimit.key;
+    element.data = lowerLimit.data;
+  } else if (rightChild.length > minDegree) {
+    const upperLimit = extractLimit(rightChild, minDegree, true);
+    element.key = upperLimit.key;
+    element.data = upperLimit.data;
+  } else {
+    joinNodes(node, elementIndex, elementIndex + 1);
+    rightChild.splice(minDegree - 1, 1);
+  }
+  return deletedData;
+};
+
+class BTree {
+
+  constructor(degree = DEFAULT_DEGREE) {
+    this.root = [empty()];
+    this.minDegree = degree;
+  }
+
+  get(key) {
+    let currNode = this.root;
+    while (currNode) {
+      const [found, i] = binarySearch(currNode, key);
+      if (found) {
+        return currNode[i].data;
+      }
+      currNode = currNode[i].child;
+    }
+    return undefined;
+  }
+
+  set(key, data) {
+    const newElement = new Element(key, data);
+    if (this.root.length === 1) {
+      this.root.unshift(newElement);
+      return this;
+    }
+    if (this.root.length === this.minDegree * 2) {
+      this.root = [empty(this.root)];
+      splitNode(this.root, 0);
+    }
+    let currNode = this.root;
+    while (true) {
+      const [found, nextNodeIndex] = binarySearch(currNode, key);
+      if (found) {
+        currNode[nextNodeIndex].data = data;
+        return this;
+      }
+      if (isLeaf(currNode)) {
+        currNode.splice(nextNodeIndex, 0, newElement);
+        return this;
+      }
+      let nextNode = currNode[nextNodeIndex].child;
+      if (nextNode.length === this.minDegree * 2) {
+        splitNode(currNode, nextNodeIndex);
+        const element = currNode[nextNodeIndex];
+        if (element.key === key) {
+          element.data = data;
+          return this;
+        }
+        if (element.key > key) {
+          nextNode = element.child;
+        }
+      }
+      currNode = nextNode;
+    }
+  }
+
+  iterator(start, finish) {
+    return inorderTraversal(start, finish, this.root);
+  }
+
+  remove(key) {
+    let currNode = this.root;
+    while (currNode) {
+      const [found, index] = binarySearch(currNode, key);
+      if (found) {
+        const deletedData = deleteElement(currNode, index, this.minDegree);
+        if (this.root.length === 1 && this.root[0].child) {
+          this.root = this.root[0].child;
+        }
+        return deletedData;
+      } else {
+        let nextNode = currNode[index].child;
+        if (nextNode && nextNode.length === this.minDegree) {
+          nextNode = growChild(currNode, index);
+        }
+        currNode = nextNode;
+      }
+    }
+    if (this.root.length === 1 && this.root[0].child) {
+      this.root = this.root[0].child;
+    }
+    return undefined;
+  }
+}
+
+module.exports = {
+  BTree
+};
--- a/test/btree.js
+++ b/test/btree.js
@@ -1,0 +1,318 @@
+'use strict';
+
+const { BTree } = require('..');
+const metatests = require('metatests');
+
+metatests.test('basic set', (test) => {
+  const bTree = new BTree(2);
+
+  test.strictSame(bTree.root, [{
+    key: undefined,
+    data: undefined,
+    child: null
+  }]);
+
+  bTree.set(40,  { city: 'Abu Dhabi' });
+
+  test.strictSame(bTree.root.length, 2);
+  test.strictSame(bTree.root[0].key, 40);
+  test.strictSame(bTree.root[0].data, { city: 'Abu Dhabi' });
+  test.strictSame(bTree.root[0].child, null);
+
+  bTree.set(30,  { city: 'Ankara'     });
+  bTree.set(50,  { city: 'Jerusalem'  });
+
+  test.strictSame(bTree.root.length, 4);
+  test.strictSame(bTree.root[0].key, 30);
+  test.strictSame(bTree.root[0].data, { city: 'Ankara' });
+  test.strictSame(bTree.root[1].key, 40);
+  test.strictSame(bTree.root[2].key, 50);
+  test.strictSame(bTree.root[3].key, undefined); // Special empty element
+  test.end();
+});
+
+metatests.test('set with spliting root', (test) => {
+  const bTree = new BTree(2);
+  bTree
+    .set(40,  { city: 'Abu Dhabi'  })
+    .set(100, { city: 'Ankara'     })
+    .set(20,  { city: 'Jerusalem'  })
+    .set(0,   { city: 'Seoul'      });
+  //        40
+  //      /    \
+  //  0 20      100
+  test.strictSame(bTree.root[0].key, 40);
+  test.strictSame(bTree.root[1].key, undefined);
+  test.strictSame(bTree.root[0].child[0].key, 0);
+  test.strictSame(bTree.root[0].child[0].data, { city: 'Seoul' });
+  test.strictSame(bTree.root[0].child[1].key, 20);
+  test.strictSame(bTree.root[0].child[2].key, undefined);
+  test.strictSame(bTree.root[1].child[0].key, 100);
+  test.strictSame(bTree.root[1].child[1].key, undefined);
+  test.end();
+});
+
+metatests.test('Insert key, which already are in', (test) => {
+  const bTree = new BTree(2);
+  bTree
+    .set('first',  { data: 'first set'  })
+    .set('second', { data: 'second set' })
+    .set('third',  { data: 'third set'  })
+    .set('fouth',  { data: 'fouth set'  })
+    .set('fouth')
+    .set('fifth',  { data: 'fifth set'  })
+    .set('first');
+
+  let result = bTree.get('first');
+  test.strictSame(result, undefined);
+
+  result = bTree.get('fouth');
+  test.strictSame(result, undefined);
+  test.end();
+});
+
+metatests.test('set with spliting internal node', (test) => {
+  const bTree = new BTree(3);
+
+  bTree
+    .set(20,  { city: 'Abu Dhabi'  })
+    .set(50,  { city: 'Ankara'     })
+    .set(100, { city: 'Jerusalem'  })
+    .set(150, { city: 'Seoul'      })
+    .set(200, { city: 'Kiev'       })
+    .set(0,   { city: 'Luxembourg' })
+    .set(160, { city: 'Minsk'      })
+    .set(180, { city: 'Oslo'       })
+    .set(170, { city: 'Stockholm'  });
+  //          100
+  //      /          \
+  //  0 20 50      150 160 170 180 200
+  bTree.set(190, { city: 'Washington' }); // splitting left node
+  //          100          170
+  //       /          \           \
+  //  0 20 50      150 160      180 190 200
+
+  test.strictSame(bTree.root.length, 3);
+  test.strictSame(bTree.root[0].key, 100);
+  test.strictSame(bTree.root[1].key, 170);
+  test.strictSame(bTree.root[0].child[0].key, 0);
+  test.strictSame(bTree.root[0].child[2].key, 50);
+  test.strictSame(bTree.root[1].child[1].key, 160);
+  test.strictSame(bTree.root[2].child[1].key, 190);
+  test.end();
+});
+
+metatests.test('get / number keys', (test) => {
+  const bTree = new BTree(3);
+
+  bTree
+    .set(20,  { city: 'Abu Dhabi'  })
+    .set(50,  { city: 'Ankara'     })
+    .set(100, { city: 'Jerusalem'  })
+    .set(150, { city: 'Seoul'      })
+    .set(200, { city: 'Kiev'       })
+    .set(0,   { city: 'Luxembourg' })
+    .set(160, { city: 'Minsk'      })
+    .set(180, { city: 'Oslo'       })
+    .set(170, { city: 'Stockholm'  });
+
+  test.strictSame(bTree.get(20),  { city: 'Abu Dhabi'  });
+  test.strictSame(bTree.get(50),  { city: 'Ankara'     });
+  test.strictSame(bTree.get(100), { city: 'Jerusalem'  });
+  test.strictSame(bTree.get(150), { city: 'Seoul'      });
+  test.strictSame(bTree.get(0),   { city: 'Luxembourg' });
+  test.strictSame(bTree.get(170), { city: 'Stockholm'  });
+  test.end();
+});
+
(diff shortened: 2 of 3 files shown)


tshemsedinov:
- Yes, it will be better to use lists, here is separate issue
- No, as of now `Object.assign` is slower then mixin
@machendos

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+          addLarger(curElement.child);
+        }
+        result.push(curElement.typle);
+        for (const curElement of curNode.elements.slice(curElementIndex + 1)) {
+          if (!curNode.leaf) {
+            addAll(curElement.child);
+          }
+          result.push(curElement.typle);
+        }
+        if (!curNode.leaf) {
```
tshemsedinov: single-line

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+          addAll(curNode.lastChild);
+        }
+        return result;
+      }
+    }
+    if (!curNode.leaf) {
```
tshemsedinov: single-line

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+    }
+    const length = curNode.elements.length;
+    if (curNode.elements[length - 1].value < finishValue && (!curNode.leaf)) {
+      addLess(curNode.lastChild);
+    }
+  };
+
+  const addLarger = function(curNode) {
+    let curElementIndex = 0;
+    for (; curElementIndex < curNode.elements.length; curElementIndex++) {
```
tshemsedinov: cache `len`

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+      const curElement = curNode.elements[curElementIndex];
```
tshemsedinov: use intermediate variables

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+      }
+    }
+    if (!curNode.leaf) {
+      addLarger(curNode.lastChild);
+    }
+  };
+
+  const addMix = function(curNode) {
+    let curElementIndex = 0;
+    for (; curElementIndex < curNode.elements.length; curElementIndex++) {
```
tshemsedinov: `len`

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+          addMix(curNode.child);
+        }
+        break;
+      }
+      if (!curNode.leaf) {
+        addLarger(curElement.child);
+      }
+      result.push(curElement.typle);
+
+      for (const curElement of curNode.elements.slice(curElementIndex + 1)) {
```
tshemsedinov: intermediate variables

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+    for (const curElement of curNode.elements) {
+      if (value < curElement.value) {
+        if (curElement.child.elements.length === this.minDegree * 2 - 1) {
+          const child = curElement.child;
+          const newLeftChild = new BTreeNode(child.leaf);
+          const newRightChild = new BTreeNode(child.leaf);
+          newRightChild.elements = child.elements.slice(this.minDegree);
+          newRightChild.lastChild = child.lastChild;
+          newLeftChild.elements = child.elements.slice(0, this.minDegree - 1);
+          newLeftChild.lastChild = child.elements[this.minDegree - 1].child;
```
tshemsedinov: put `child.elements[this.minDegree - 1]` to intermediate

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+    // Must to add in lastChild
+    if (needChangedNodeFlag) {
+      if (curNode.lastChild.elements.length === this.minDegree * 2 - 1) {
+        const child = curNode.lastChild;
+        const newLeftChild = new BTreeNode(child.leaf);
+        const newLastChild = new BTreeNode(child.leaf);
+        newLastChild.elements = child.elements.slice(this.minDegree);
+        newLastChild.lastChild = child.lastChild;
+        newLeftChild.elements = child.elements.slice(0, this.minDegree - 1);
+        newLeftChild.lastChild = child.elements[this.minDegree - 1].child;
```
tshemsedinov: put `child.elements[this.minDegree - 1]` to variable

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+        curNode.lastChild = newLastChild;
+        curNode = value < median.value ? newLastChild : newLastChild;
+        break;
+      }
+      curNode = curNode.lastChild;
+    }
+  }
+  // Now in curNode is node (leaf), in which we need to insert an element
+  let curElementIndex = 0;
+  for (; curElementIndex < curNode.elements.length; curElementIndex++) {
```
tshemsedinov: cache `len`

tshemsedinov:
@machendos here is issue for lists https://github.com/metarhia/common/issues/160

aqrln:
> Perhaps it would be better to remake this to a one-linked list? Since we never need a quick access to an item by index. And when you search for an element, you still look at the part of the array that precedes the desired one.

You won't be able to perform binary search within the node then.

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+
+const DEFAULT_DEGREE = 6; // min degree of b-tree.
+// All vertices except the root have [degree ... 2 * degree] child nodes
+
+function BTree(degree = DEFAULT_DEGREE) {
+  this.root = new BTreeNode();
+  this.minDegree = degree;
+}
+
+function BTreeNode(leaf = true) {
```
lundibundi: No need to add BTree, this is already scoped inside BTree file. So just `Node` is fine.
Also same as for Btree, use es6 classes, please.

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+  this.elements = [];
+  this.leaf = leaf;
+  this.lastChild = null;
+}
+
+function Element(value, typle) {
+  this.value = value;
+  this.typle = typle;
```
lundibundi: s/typle/tuple ?

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
+'use strict';
+
+const DEFAULT_DEGREE = 6; // min degree of b-tree.
+// All vertices except the root have [degree ... 2 * degree] child nodes
+
+function BTree(degree = DEFAULT_DEGREE) {
```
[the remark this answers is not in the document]
tshemsedinov: We have no style recommendations about prototype/class usage in Metarhia libs and prototypes are still a bit faster then classes as I tested a few days ago for latest node build but I think it's a time to use classes at least for new code and we need performance comparison and explanation for all places where we fallback to prototypes for performance reasons.

@lundibundi @machendos 

/cc @aqrln

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+  this.minDegree = degree;
+}
+
+function BTreeNode(leaf = true) {
+  this.elements = [];
+  this.leaf = leaf;
+  this.lastChild = null;
+}
+
+function Element(value, typle) {
```
lundibundi: You already have `leaf` property in `BTreeNode`, why not use it and store inherent data in elements?

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+  this.child = null;
+}
+
+BTree.prototype.getEqual = function(value) {
+  let curNode = this.root;
+  while (true) {
+    let needChangedNodeFlag = true;
+    for (const curElement of curNode.elements) {
+      if (value <= curElement.value) {
+        if (curElement.value === value) return curElement.typle;
```
lundibundi: Why compare for equality twice? (here and above)

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+        if (curNode.leaf) return null;
```
lundibundi: Why check for `leaf` twice (here and blow on line 36)?

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+        curNode = curElement.child;
+        needChangedNodeFlag = false;
+        break;
+      }
+    }
+    if (curNode.leaf) return null;
+    if (needChangedNodeFlag) {
+      curNode = curNode.lastChild;
+    }
+  }
```
lundibundi: Whole loop looks strange and could be rewritten to avoid unnecessary checks and flags as well as `break` usage.
Also, it will be better to write separate `find` function to find element equals-or-before given in a list and use it here.

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+};
+
+BTree.prototype.getLarger = function(value) {
+  const result = [];
+  const addAll = function(curNode) {
+    if (curNode.leaf) {
```
lundibundi: Weird `if`, this is a simple constant-propagation optimization, I don't think it's worth 'uglifying' code this much just for it.

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+
+function BTree(degree = DEFAULT_DEGREE) {
+  this.root = new BTreeNode();
+  this.minDegree = degree;
+}
+
+function BTreeNode(leaf = true) {
+  this.elements = [];
+  this.leaf = leaf;
+  this.lastChild = null;
```
lundibundi: Why is `lastChild` stored separately?

lib/b-tree.js
```diff
@@ -0,0 +1,318 @@
...
+      addAll(curNode.lastChild);
+    }
+  };
+
+  const addLarger = function(curNode) {
+    let curElementIndex = 0;
+    for (; curElementIndex < curNode.elements.length; curElementIndex++) {
+      const curElement = curNode.elements[curElementIndex];
+      if (curElement.value <= value) {
+        continue;
```
lundibundi: Why `continue`, elements are sorted you will not get any more <= elements. You can just `break` at this point.

lib/b-tree.js
machendos: These are different blocks. The first block for nodes with values not less than the specified, and on line 36 for those with a value less than

lib/b-tree.js
machendos: In each element of the node store a reference to a node with elements less than it. Therefore, somewhere in the node, we still need to store a reference to a node with elements larger than the last element in the node. I decided to store it in a separate field.

lib/b-tree.js
machendos: ``leaf`` is just a flag that informs: is the current node a leaf or not? I dont quite understand how we can store data there?

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+  const len = node.length;
+  const newLeftNode = node.splice(0, len / 2 - 1); // First half
+  const mid = node.shift();
+  newLeftNode.push(empty(mid.less));
+  mid.less = newLeftNode;
+  parent.splice(index, 0, mid);
+  return parent;
+};
+
+const isLeaf = node => !node[0].less;
```
lundibundi: Is this correct? The element may have a subtree under any of the children AFAIK, isn't it? i.e.
```
[1, 5, 15, 30]
       ↙
     [7, 8]  
```

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+  let end = node.length - 1;
+  while (start <= end) {
+    const i = (start + end) >> 1;
+    const item = node[i];
+    const itemValue = item.value;
+    if (value > itemValue) start = i + 1;
+    else if (!(value >= itemValue)) end = i - 1;
+    // undefined must be bigger then other value
+    // (!(1 >= undefined)) -> true, but just (1 < undefined) -> false
+    else return [true, item.tuple];
```
lundibundi: I'd suggest returning at least an item itself (I assume you optimized it this way). Though I'd prefer to always return an integer (index) from this function this will make it both easier to understand and easier to optimize for the compiler (less type diversity, only integers).

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+  return [false, start];
+};
+
+function* inorderTraversal(start, finish, currNode) {
+  let [found, startIndex] = binarySearch(currNode, start);
+  if (found) startIndex++;
```
lundibundi: Your binsearch returns an `item.tuple` upon found === true, therefore you are incrementing an `item.tuple` here, are you sure this is correct?

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+  let [_, finishIndex] = binarySearch(currNode, finish);
+  if (finish === undefined)
```
lundibundi: This check should go *before* the search (binsearch of finishIndex) and the search will be in the else branch.

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+      const [found, i] = binarySearch(currNode, value);
+      if (found)
+        return i;
+      currNode = currNode[i].less;
+    }
+    return null;
+  }
+
+  insert(value, tuple) {
+    const alreadyHaveErrMes = `value ${value} already is in the tree`;
```
lundibundi: Move this to the place it is used in, you don't have to create it always. It's better to copy the error here (or make a function that will return this error) that create it every time we enter this function.

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+    const newElement = new Element(value, tuple);
+    if (this.root.length === 1)
+      return (this.root.unshift(newElement), this);
```
lundibundi: Please, don't use 'coma' unless absolutely necessary.

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+const DEFAULT_DEGREE = 6; // min degree of b-tree.
+// All vertices except the root have [degree ... 2 * degree] child nodes
+// And [degree - 1 ... 2 * degree - 1] + 1 "empty" elements
+
+class Element {
+  constructor(value, tuple, less = null) {
+    this.value = value;
+    this.tuple = tuple;
+    this.less = less;
+  }
```
lundibundi: Please rename `less` to `child` or smth. Imo less is confusing.

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+  constructor(degree = DEFAULT_DEGREE) {
+    this.root = [empty()];
+    this.minDegree = degree;
+  }
+
+  get(value) {
+    let currNode = this.root;
+    while (currNode) {
+      const [found, i] = binarySearch(currNode, value);
+      if (found)
```
lundibundi: We always use braces, add them here and in other places. OR you can go with one liner: 
`if (found) return i;`

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+        if (element.value === value)
+          throw new Error(alreadyHaveErrMes);
+        if (element.value > value)
+          nextNode = element.less;
+      }
+      currNode = nextNode;
+    }
+  }
+
+  iterator(start, finish = undefined) {
```
lundibundi: `start` can also be undefined, you will then traverse the whole tree.

lib/btree.js
```diff
@@ -0,0 +1,115 @@
...
+
+const binarySearch = (node, value) => {
+  let start = 0;
+  let end = node.length - 1;
+  while (start <= end) {
+    const i = (start + end) >> 1;
+    const item = node[i];
+    const itemValue = item.value;
+    if (value > itemValue) start = i + 1;
+    else if (!(value >= itemValue)) end = i - 1;
```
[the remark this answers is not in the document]
machendos: Each node has an element with ``value = undefined`` that is not less than all other possible values in the remaining nodes. This special element is created by a function ``empty``. I wrote about this on gitlab

lib/btree.js
machendos: In the b-tree, all elements from the node have children, or none. Therefore, we can only check the first element, and be sure that the rest of the elements in this node have a similar situation

lib/btree.js
machendos: Yes exactly. I do not know how the tests did not reveal this. Then probably, like you said, it's better to always return the index from binary search, instead of ``item.tuple``

lib/btree.js
machendos: I remembered that if we didn`t pass the value into the argument of the function, then it would become an ``undefined``. Therefore, we can write just this: -
``iterator(start, finish) {}``

lundibundi: Yes, if you don't need specific values (not undefined) for defaults you can just remove the defaults.

lib/btree.js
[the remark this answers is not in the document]
lundibundi: Ok, then change the comment to something similar to
```
// the biggest element in the node has value === undefined, therefore,
// in order to get true for it here using reverse comparison.
```
And put it above the line.

lib/btree.js
```diff
@@ -0,0 +1,134 @@
...
+  constructor(value, tuple, child = null) {
+    this.value = value;
+    this.tuple = tuple;
+    this.child = child;
+  }
+}
+
+const empty = child => new Element(undefined, null, child);
+
+const destroyNode = (parent, index) => {
```
lundibundi: Rename it to smth like `splitNode`.

lib/btree.js
```diff
@@ -0,0 +1,134 @@
...
+  if (found) startIndex++;
+  let finishIndex;
+  if (finish === undefined) {
+    finishIndex = currNode.length - 1;
+  } else {
+    let _; // We need to initialize the unnecessary variable to use
+    // the destructive assignment in the next line
+    [_, finishIndex] = binarySearch(currNode, finish);
+  }
+  for (let currIndex = startIndex; currIndex <= finishIndex; currIndex++) {
```
lundibundi: What do you think of
```
if (isLeaf(currNode)) {
  for (let currIndex = startIndex; currIndex < finishIndex; currIndex++) {  
    yield currNode[currIndex].tuple;   
  }
} else {
  for (let currIndex = startIndex; currIndex < finishIndex; currIndex++) {
    const currElement = currNode[currIndex];
    yield* inorderTraversal(start, finish, currElement.child);  
    yield currElement.tuple;   
  }
  yield* inorderTraversal(start, finish, currNode[finishIndex].child);  
}
```
?

lib/btree.js
```diff
@@ -0,0 +1,134 @@
...
+
+const DEFAULT_DEGREE = 6; // min degree of b-tree.
+// All vertices except the root have [degree ... 2 * degree] child nodes
+// And [degree - 1 ... 2 * degree - 1] + 1 "empty" elements
+
+class Element {
+  constructor(value, tuple, child = null) {
+    this.value = value;
+    this.tuple = tuple;
+    this.child = child;
```
lundibundi: :thinking: it may even be `children`, wdyt?

lib/btree.js
```diff
@@ -0,0 +1,134 @@
+'use strict';
+
+const DEFAULT_DEGREE = 6; // min degree of b-tree.
+// All vertices except the root have [degree ... 2 * degree] child nodes
+// And [degree - 1 ... 2 * degree - 1] + 1 "empty" elements
+
+class Element {
+  constructor(value, tuple, child = null) {
+    this.value = value;
+    this.tuple = tuple;
```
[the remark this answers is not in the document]
machendos: I agree, I think it will be more correct for a general purpose btree

lib/btree.js
machendos: I don't think that it will be more correct, we have here only a one ref to one child and not to several

lib/btree.js
machendos: Yes, I also wanted to write this way, and I thought that it would faster. But I was afraid that there is a copy-paste. If this is not very critical, I implement this variant

lib/btree.js
machendos: No, this is not a bug. This function split one child into two. The left half will become the child of the new element that will be inserted before element with position ``index`` in the node. And right half remains a child of its element

lib/btree.js
```diff
@@ -0,0 +1,129 @@
...
+      const [found, i] = binarySearch(currNode, value);
+      if (found) {
+        return currNode[i].data;
+      }
+      currNode = currNode[i].child;
+    }
+    return null;
+  }
+
+  insert(value, data) {
```
lundibundi: Could you also add a `replace` boolean flag parameter to just replace the value if it already exists in the tree?

machendos: I also thought about the flag ``include`` in the function  ``iterator`` which shows: the iterator includes or not the edges of the segment in the result. It makes sense?

machendos: It seems to me most likely it will be necessary to include either both borders, or not include a single one. So maybe we can just leave one flag that will inform about both borders

lundibundi: Well, I'm not so sure, I think we can discuss this later. In the end, it's easy to add such a flag.

machendos: OK

lib/btree.js
```diff
@@ -0,0 +1,240 @@
...
+
+class Element {
+  constructor(value, data, child = null) {
+    this.value = value;
+    this.data = data;
+    this.child = child;
+  }
+}
+
+const empty = child => new Element(undefined, null, child);
```
lundibundi: I think `undefined` is better for `data` here.

lib/btree.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+  }
+
+  iterator(start, finish) {
+    return inorderTraversal(start, finish, this.root);
+  }
+
+  remove(value) {
+    let currNode = this.root;
+    while (!isLeaf(currNode)) {
```
lundibundi: Why don't you change this to `while (currNode)` (or !== null / !== undefined) and then you won't need the code on L232-L235. If that's because delete doesn't work for Leafs then I think adding a `!isLeaf` check in the last else branch on L152 might fix that.

lib/btree.js
```diff
@@ -0,0 +1,240 @@
...
+          nextNode = growChild(currNode, index);
+        }
+        currNode = nextNode;
+      }
+    }
+    if (this.root.length === 1) this.root = this.root[0].child;
+    const [found, index] = binarySearch(currNode, value);
+    if (found) {
+      return currNode.splice(index, 1)[0].data;
+    } else return null;
```
lundibundi: If the comment about removing this code is not applicable then you have to fix the formatting of this, we either have
* both branches with braces or
* both branches without braces

lib/btree.js
```diff
@@ -0,0 +1,240 @@
...
+  return test;
+};
+
+const deleteElement = (node, elementIndex, minDegree) => {
+  const element = node[elementIndex];
+  const deletedData = element.data;
+  const leftChild = element.child;
+  const rightChild = node[elementIndex + 1].child;
+  if (leftChild.length > minDegree) {
+    const lowerLimit = extractLowerLimit(leftChild, minDegree);
```
lundibundi: I'm not sure, but shouldn't you search for the highest element in the left child and lowest in the right? (here it is reversed) Or maybe it's the function name that are confusing?

lib/btree.js
```diff
@@ -0,0 +1,240 @@
...
+    let nextNode = currNode[0].child;
+    if (nextNode.length === minDegree) {
+      nextNode = growChild(currNode, 0);
+    }
+    currNode = nextNode;
+  }
+  return currNode.shift();
+};
+
+const extractLowerLimit = (node, minDegree) => {
```
lundibundi: I think this and the function above can be merged via a single `i` argument (or bool argument and set i in the beginnig of the function) that will be 0 if Upper or currNode.length - 1 if Lower. And simple if check in the end.

lib/btree.js
```diff
@@ -43,7 +43,9 @@ const binarySearch = (node, value) => {
 };
 
 function* inorderTraversal(start, finish, currNode) {
-  let [found, startIndex] = binarySearch(currNode, start);
+  const binarySearchResult = binarySearch(currNode, start);
+  const found = binarySearchResult[0];
```
lundibundi: There is no need to have this as a variable, you only use it once. Just inline it.

lib/btree.js
machendos: I thought so: the upper limit of an element is the smallest element in the right node, and the lower limit is the smallest element in the left node. Is it more correct to rename?

lib/btree.js
```diff
@@ -0,0 +1,240 @@
...
+        parent[childIndex - 1].data,
+        leftNeighbor[leftNeighbor.length - 1].child
+      );
+      leftNeighbor[leftNeighbor.length - 1].child = extractedElement.child;
+      extractedElement.child = leftNeighbor;
+      parent.splice(childIndex - 1, 1, extractedElement);
+      node.unshift(insertedElement);
+      return node;
+    }
+  } else if (childIndex < parent.length - 1) { // If we have right neighbor
```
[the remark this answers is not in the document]
machendos: Yes, it's wrong. Now it's fixed. Now this happens when we have the index === 0, or if the right neighbor exists, but its length is equal to the minimum index of the b-tree

lib/btree.js
```diff
@@ -0,0 +1,240 @@
...
+      const extractedElement = rightNeighbor.shift();
+      const insertedElement = new Element(
+        parentElement.value,
+        parentElement.data,
+        node[node.length - 1].child
+      );
+      node[node.length - 1].child = extractedElement.child;
+      extractedElement.child = node;
+      parent.splice(childIndex, 1, extractedElement);
+      node.splice(node.length - 1, 0, insertedElement);
```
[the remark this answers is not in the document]
machendos: Not really, we move the last item, and insert instead the ``insertedElement``

lib/btree.js
machendos: The index is always the same in the ``extractUpperLimit``, it is equal to 0, but in the ``extractLoverLimit`` it changes every iteration, so, every time when we use the index, it will be necessary to check which limit we need. Is it good?

lib/btree.js
lundibundi: The right node contains all elements bigger than this one, and you find the smallest element in that node, therefore, it'd be better to call it by exactly what it does - finds the smallest element (and same goes for the other function but reverted).

lib/btree.js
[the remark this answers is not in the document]
lundibundi: Oh, you insert at a place of the last element and move it to the next element. I got it.

test/btree.js
```diff
@@ -0,0 +1,207 @@
...
+    .insert(18,  { city: 'Minsk'      })
+    .insert(135, { city: 'Oslo'       })
+    .insert(50,  { city: 'Stockholm'  })
+    .insert(25,  { city: 'Washington' })
+    .insert(62,  { city: 'Lima'       })
+    .insert(64,  { city: 'Beijing'    });
+
+  bTree.remove(40);
+  let res = bTree.get(40); // Remove root
+  test.strictSame(res, null);
```
lundibundi: If must be not null, but `undefined` here. That's what `undefined` means - no such value.

test/btree.js
```diff
@@ -0,0 +1,207 @@
...
+    .insert('bbb',  2)
+    .insert('ccca', 3)
+    .insert('cccz', 4)
+    .insert('ddd',  5)
+    .insert('eee',  6)
+    .insert('fff',  7);
+  let res = [...bTree.iterator('bbb', 'eee')].join('');
+  test.strictSame(res, '345');
+
+  res = [...bTree.iterator(undefined)].join('');
```
lundibundi: No need for explicit undefined.

lundibundi:
Also, could you add a test that will cover the following lines at least once:
image
? (image from coverage report, you can get it via `nyc --reporter=html npm test`, nyc must be installed globally via npm)

lib/btree.js
```diff
@@ -0,0 +1,245 @@
...
+      return node;
+    }
+    return joinNodes(parent, childIndex, childIndex + 1);
+  }
+  return joinNodes(parent, childIndex - 1, childIndex);
```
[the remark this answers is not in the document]
machendos: Not really, this code works when we don't have a left neighbor, and when the length of the right neighbor is equal to bTree.minDegree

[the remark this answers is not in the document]
machendos: Also, if i correctly understand, how ``nyc`` works, you can see that tests enter this block, and work out correctly:
image_2018-09-06_19-28-23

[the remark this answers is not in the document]
lundibundi: :thinking: According to your decision tree:
If we go right->right then we don't have left or right neighbor but we 'join out node with left node'?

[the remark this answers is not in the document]
machendos: Not all routes in this decision tree are achievable. We can't go to the right ->right, since at least one neighbor we must have, this is embedded in the rules of the b-tree. If we go to the right in the beginning, then we must go to the left. I did not put this in a separate block, because it seemed to me that too much code would be repeated, or it would be necessary to create additional functions. What do you think, which option is acceptable?

machendos:
@lundibundi. Question about ``replace`` flag in ``insert`` method. It seems to me that it is more logical make this flag in constructor of B tree. it will mean whether a tree can contain the same values. And then flag in function not required.

lib/btree.js
```diff
@@ -0,0 +1,238 @@
...
+};
+
+const extractLimit = (
+  node,
+  minDegree,
+  upper // Boolean: true - upperLimit, false - loverLimit
```
lundibundi: s/loverLimit/lowerLimit

lundibundi:
@machendos I'm not sure I understand how is it feasible to allow multiple elements with the same value in a BTree? we would then need to make `get` always return an array, `delete` always accept a value/data pair instead of just value, it will complicate the process of searching for a given value etc.
Also, we could actually rename `insert` -> `set` and make it always insert or replace without any flag. This will be consistent with other JS data structures. Wdyt?

lib/btree.js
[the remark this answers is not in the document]
machendos: We can get to the block ``if`` on L97 from two places: from if on L80 and L82. Accordingly, from L80 we can't get to L114, but for L82 we can. That is, if we have a left neighbor, but it is too small to extract an element from it, we will get to L97, and then, if we have a right neighbor, we'll get on L114.

machendos:
@lundibundi I was wrong, do not pay attention. It was a bad proposition. About, ``set`` okay, a good idea. 
Сan also make a similar safe tree, which does not overwrite the old values, but somehow reacts to the attempt of overwriting?

lib/btree.js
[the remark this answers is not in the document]
lundibundi: You mean ' if we have a right neighbor, we'll get on L114.' if we **don't** have a right neighbor?

lundibundi:
I'm not sure for the API of that, I think we can implement that later if we need it. Though if you have a good grasp of how it should be done you can do it in this PR as well.

lib/btree.js
```diff
@@ -0,0 +1,238 @@
+'use strict';
+
+const DEFAULT_DEGREE = 6; // min degree of b-tree.
+// All vertices except the root have [degree ... 2 * degree] child nodes
+// And [degree - 1 ... 2 * degree - 1] + 1 "empty" elements
+
+class Element {
+  constructor(value, data, child = null) {
```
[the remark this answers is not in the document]
machendos: ``value`` - is number or string on which the tree will be built. ``data`` is data of any kind associated with these ``value``: objects, arrays, etc.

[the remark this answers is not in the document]
machendos: I change ``value`` into ``key``

lundibundi:
@machendos could you squash all of the commits and rebase this on master to simplify finishing this PR (as there are way too many commits to use them efficiently to review this)?

test/btree.js
```diff
@@ -0,0 +1,318 @@
...
+  test.strictSame(res, 'ABCDEFGHIJKLMNOPQR');
+
+  res = [...bTree.iterator(100, 0)].join(''); // Empty
+  test.strictSame(res, '');
+
+  res = [...bTree.iterator(-200, -100)].join(''); // Empty
+  test.strictSame(res, '');
+
+  res = [...bTree.iterator(20, 50)].join('');
+  test.strictSame(res, 'EFGH');
```
belochub: Why does not this include 'D' if it was added as an element with key 20?

lib/btree.js
```diff
@@ -0,0 +1,247 @@
...
+
+const binarySearch = (node, key) => {
+  let start = 0;
+  let end = node.length - 1;
+  while (start <= end) {
+    const i = (start + end) >> 1;
+    const item = node[i];
+    const itemValue = item.key;
+    if (key > itemValue) start = i + 1;
+    // the biggest element in the node has key === undefined, therefore,
```
belochub: How about just adding the check if `key === undefined` instead of using implicit type conversion in strange looking value checks like the one below?

test/btree.js
```diff
@@ -0,0 +1,318 @@
...
+  bTree.remove(135);
+
+  res = bTree.remove(0); // Try to remove nonexistent node
+  test.strictSame(res, undefined);
+
+  bTree.remove(25);
+  bTree.remove(30);
+  bTree.remove(20);
+  bTree.remove(110);
+  // Now bTree is emty
```
belochub: Typo in the word 'empty' here.

lib/btree.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+    return joinNodes(parent, childIndex, childIndex + 1);
+  }
+  return joinNodes(parent, childIndex - 1, childIndex);
+};
+
+const extractLimit = (
+  node,
+  minDegree,
+  upper // Boolean: true - upperLimit, false - lowerLimit
```
lundibundi: Use new comment style here, please.

test/btree.js
lundibundi: I thought we agreed to use [a, b) sort of range?

machendos: I don't remember when we agreed this, but okay, I'll do so

lib/btree.js
```diff
@@ -0,0 +1,245 @@
...
+
+const binarySearch = (node, key) => {
+  let start = 0;
+  let end = node.length - 1;
+  while (start <= end) {
+    const i = (start + end) >> 1;
+    const itemKey = node[i].key;
+    if (key > itemKey) start = i + 1;
+    else if (itemKey === undefined) end = i - 1;
+    else if (key < itemKey) end = i - 1;
```
belochub: Merge this and previous lines, please.

lundibundi:
@machendos could you rebase this on master and we will then merge this?

belochub:
Landed in https://github.com/metarhia/common/commit/1ac4d753b2ee189b4151bf738f51bb403a949ec4.

Issue #161, opened 2018-08-09
Infinite integer, class Id


tshemsedinov:
We need class Id for infinite integer, it will be used in globalstorage, file storage memory datasets to represent entity identifier, etc. Here is a branch with code stub: https://github.com/metarhia/common/tree/infinite-integer

SemenchenkoVitaliy:
I think there should be functions like
next() - increments id value by 1 and returns it
set(n) - sets id value to n

aqrln:
`BigInt` is now built into the language. Officially it's still stage 3, but it's going to become a part of the standard quite soon, and it has already been around in V8 and Node.js for some time:

```
$ node --version
v10.8.0
$ node
> 10000n ** 10000n
[...]

PR #162, opened 2018-08-09
Infinite integer

SemenchenkoVitaliy:
Implement Id class
    
Closes: https://github.com/metarhia/common/issues/161

### DIFF
--- a/lib/id.js
+++ b/lib/id.js
@@ -147,6 +147,29 @@ const pathToId = path => {
   return parseInt(hex, 16);
 };
 
+class Id {
+  constructor(
+    n // Initial value
+    // string or number or Array of Number, Uint32Array
+  ) {
+    this.value = n;
+  }
+  toChunks() {
+  }
+  toPath() {
+  }
+  toNumber() {
+  }
+  toString() {
+  }
+  [Symbol.toPrimitive](hint) {
+    if (hint === 'number') {
+      return this.value;
+    }
+    return null;
+  }
+}
+
 module.exports = {
   generateKey,
   generateGUID,
@@ -158,5 +181,6 @@ module.exports = {
   generateStorageKey,
   idToChunks,
   idToPath,
-  pathToId
+  pathToId,
+  Id,
 };
--- a/lib/id.js
+++ b/lib/id.js
@@ -139,7 +139,7 @@ const idToPath = id => {
 
 const pathToId = path => {
   const chunks = path.split('/');
-  let hex = '0x';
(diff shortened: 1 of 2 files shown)


SemenchenkoVitaliy:
Btw, is there a better way of testing class methods? And is it ok to take repeated data outside of test functions?

lib/id.js
```diff
@@ -139,14 +139,96 @@ const idToPath = id => {
...
   }
   return parseInt(hex, 16);
 };
 
+class Id {
+  constructor(
+    n // Initial value
+    // string or number or Array of Number, Uint32Array
+  ) {
+    if (typeof(n) === 'number')
```
tshemsedinov: If we use `else if` we need block statements

SemenchenkoVitaliy:
Also, we need to update .eslintrc.yml to be able to use BigInt and to break tests

lib/id.js
```diff
@@ -139,14 +139,96 @@ const idToPath = id => {
...
+      this.value = n;
+    else if (typeof(n) === 'string')
+      this.value = parseInt(n, 10) || 0;
+    else if (Array.isArray(n) &&
```
tshemsedinov: It will be better to put conditions in single line

lib/id.js
```diff
@@ -139,14 +139,96 @@ const idToPath = id => {
...
 
+class Id {
+  constructor(
+    n // Initial value
+    // string or number or Array of Number, Uint32Array
+  ) {
+    if (typeof(n) === 'number')
+      this.value = n;
+    else if (typeof(n) === 'string')
+      this.value = parseInt(n, 10) || 0;
```
tshemsedinov: Not `0` better `NaN` or `undefined`

lib/id.js
```diff
@@ -139,14 +139,96 @@ const idToPath = id => {
...
+  }
+
+  toString(
+    // Get current id value
+    // Returns: string
+  ) {
+    return this.value + '';
+  }
+
+  next(
```
tshemsedinov: Need no `next`

lib/id.js
```diff
@@ -139,14 +139,96 @@ const idToPath = id => {
...
+    // Increment id value by 1
+    // Returns: number
+  ) {
+    return ++this.value;
+  }
+
+  set(
```
tshemsedinov: Need no `set`

tshemsedinov:
Closed without landing, here is explanation: https://github.com/metarhia/common/issues/161

PR #166, opened 2018-08-12
Replace incorrect result value in network.ipToInt test

lundibundi:
Previous testing framework tested this incorrectly (NaN was equal to null in declarative tests).

Also if we really want `ipToInt` to return `null` in such cases I will then change the code. Though I think it's correct to return NaN because it's supposed to convert to a number, which NaN is.

Refs: https://github.com/metarhia/metatests/pull/64


lundibundi:
Landed in 0cdf0495a5748c6a1e4c84ca87685259296d8dd8.

PR #171, opened 2018-08-19
Implement mergeObjects to merge with user function

lundibundi:
This will allow to avoid writing a lot of boilerplate code when you need
to merge two (or more) objects using specific resolution mechanism for
values under the same keys.

### DIFF
--- a/lib/data.js
+++ b/lib/data.js
@@ -202,6 +202,30 @@ const merge = (
   return [...unique];
 };
 
+const mergeObjects = (
+  // Merge multiple object with merger
+  merger,
+  ...objs // array of objects
+  // Returns: object
+) => {
+  const keys = new Set();
+  for (const obj of objs) {
+    for (const key in obj) {
+      keys.add(key);
+    }
+  }
+
+  const result = {};
+  for (const key of keys) {
+    const args = new Array(objs.length);
+    for (let i = 0; i < objs.length; ++i) {
+      args[i] = objs[i][key];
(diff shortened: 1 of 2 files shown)


lib/data.js
```diff
@@ -202,6 +202,30 @@ const merge = (
...
+    for (const key in obj) {
+      keys.add(key);
+    }
+  }
+
+  const result = {};
+  for (const key of keys) {
+    const args = new Array(objs.length);
+    for (let i = 0; i < objs.length; ++i) {
+      args[i] = objs[i][key];
```
tshemsedinov: What if `objs[i][key]` will be `undefined`?

tshemsedinov: Should all objects have all keys?

lundibundi: First one - yeah will fix it
Second - no, it will just be undefined, that is expected, the merger function has to deal with that.

lundibundi:
How can we achieve this functionality with common.mixin and object.assign?
For example as in tests, I need to sum the values in the both (potentially many) objects where some of them may be missing.

tshemsedinov:
@lundibundi fine, you can merge this

lundibundi:
@tshemsedinov could you please explicitly Approve this, so I can include you in the list of reviewers?

lundibundi:
Landed in b00b34bc9f2a98830a3bafe1f1aa2b2b77fb12c1.

Issue #169, opened 2018-08-19
Some of the tests are not being run


lundibundi:
Currently, some of the tests are not running. For example, if you change https://github.com/metarhia/common/blob/master/test/zip.js#L21 to something like `test.strictSame(111, 42)` and then run `npm test` it will not fail.
On the other hand if you change test/cache.js it will properly fail.

@JuliaGerasymenko @bugagashenkj or someone else, if you have time please check.

tshemsedinov:
I think it was fixed here https://github.com/metarhia/metatests/pull/60
@lundibundi

lundibundi:
@tshemsedinov I've locally changed version to ^0.1.12 but it still didn't run that test in zip.js.

lundibundi:
On a side note, I tried it with this https://github.com/metarhia/metatests/pull/66 PR and everything worked. I also found a bug in tests. I'll add a PR soon.

tshemsedinov:
As of now `metatests 0.1.12` gives following output:
```
 Test failed: zip with no elements
        Type: strictSame
    Subtests: 1
    Expected: 42
      Actual: 111
     Message: undefined
       Stack: Error
            Test.api.metatests.test [as execute] (/test/zip.js:21:8)
            next (/node_modules/metatests/lib/test.js:224:10)
            Object.test.on [as done] (/node_modules/metatests/lib/test.js:222:7)
            Test.end (/node_modules/metatests/lib/test.js:142:39)
            Test.api.metatests.test [as execute] (/test/zip.js:16:8)
            next (/node_modules/metatests/lib/test.js:224:10)
            Object.test.on [as done] (/node_modules/metatests/lib/test.js:222:7)
            Test.end (/node_modules/metatests/lib/test.js:142:39)
            Test.api.metatests.test [as execute] (/test/sort.js:77:8)

Test finished with following results
     Modules: 10
     Targets: 150
   Functions: 50
      Values: 1
     Classes: 0
     Methods: 0
       Tests: 529
      Errors: 1
      Result: FAILED
```
@lundibundi

PR #172, opened 2018-08-19
Fix Enum.NaE test (add test.end())

lundibundi:
As I said here https://github.com/metarhia/common/issues/169#issuecomment-414145340, this test wasn't working.


lundibundi:
Landed in 7ebab584d1e5149dda8032a59a8faa06839d0099.

PR #176, opened 2018-08-28
Refactor loop variables

JuliaGerasymenko:


lundibundi:
lib/mp.js
10:  for (name in iface) {
25:  for (name in iface) {

Easy search `ag 'for \(' .`
Harder: `ag 'for \((?!(let|const))' .`

(ag - the silver searcher may be replaced with `grep -n`)

lib/data.js
```diff
@@ -122,10 +118,9 @@ const getByPath = (
 ) => {
   const path = dataPath.split('.');
   let obj = data;
-  let i, len, next, prop;
-  for (i = 0, len = path.length; i < len; i++) {
-    prop = path[i];
-    next = obj[prop];
+  for (let i = 0, len = path.length; i < len; i++) {
```
lundibundi: (No need to change this, just a topic to discuss)
I don't this that we should cache the length where we don't mutate the array either, though no strong opinion / not enough info on the impact.

/cc @aqrln

JuliaGerasymenko: Did you mean to do that one?

let i, len;
for (i = 0, len = path.length; i < len; i++) {
  const prop = path[i];
  const next = obj[prop];

JuliaGerasymenko:
About lib/mp.js. Sorry I forgot to save

lib/data.js
tshemsedinov: I'd prefer
```js
const len = path.length;
for (let i = 0; i < len; i++) {
```

lundibundi: No, I meant to just use
`for (let i = 0; i < path.length; i++) {`

My comment was a discussion topic, not a change request/suggestion, hence this comment '(No need to change this, just a topic to discuss)'.

aqrln: @lundibundi @tshemsedinov both

```js
const len = path.length;
for (let i = 0; i < len; i++) {
```

and 

```js
for (let i = 0; i < path.length; i++) {
```

would be compiled to identical machine code due to loop-invariant code motion optimization.

If the compiler is unable to infer that `path.length` doesn't change for some reason, caching it manually won't help, because chances are the compiler won't be able to eliminate bounds check for the exact same reason, so `path[i]` will access `path.length` under the hood anyway. 

That said, `path.length` is basically one `mov` instruction in optimized code (thanks to V8's ICs/maps/hidden objects/whatever you call them), so it's rather a matter of code style than performance, one way or the other.

JuliaGerasymenko:
Landed in 9df8028

PR #180, opened 2018-08-31
Comments refactor

SemenchenkoVitaliy:
Refactor comments to achieve comments style consistency and fix grammar mistakes.
Also, I want to notice that `deprecate` and `alias` can not be called without function, although it is said that `fn` parameter is optional. Should I remove this comments or open new PR and fix this?

### DIFF
--- a/common.js
+++ b/common.js
@@ -6,14 +6,14 @@ const submodules = [
   'array', // Arrays manipulations
   'data', // Data structures manipulations
   'strings', // Strings utilities
-  'time', // Data and Time functions
+  'time', // Date and Time functions
   'fp', // Functional programming
   'oop', // Object-oriented programming
   'callbacks', // Callback utilities
   'events', // Events and emitter
   'units', // Units conversion
   'network', // Network utilities
-  'id', // Kyes and identifiers
+  'id', // Keys and identifiers
   'sort', // Sort compare functions
   'cache', // Cache (enhanced Map)
   'mp', // Metaprogramming
--- a/lib/callbacks.js
+++ b/lib/callbacks.js
@@ -50,7 +50,7 @@ const once = (
 
 const unsafeCallback = (
   // Extract callback function
-  // It's unsafe: may return null, allow multiple calls
+  // It's unsafe: may return null, allows multiple calls
   args // array, arguments
   // Returns: function, callback or null
   // Hint: previous name: `common.cbUnsafe` (deprecated)
@@ -62,7 +62,7 @@ const unsafeCallback = (
 };
 
 const safeCallback = (
-  // Extracts callback and return common.emptiness if no callback
+  // Extract callback and return common.emptiness if there is no callback
   args // array, arguments
   // Returns: function, wrapped callback
   // Hint: previous name: `cbExtract` (deprecated)
@@ -73,7 +73,7 @@ const safeCallback = (
 };
 
 const requiredCallback = (
-  // Extracts callback and throw if no callback
+  // Extract callback and throw TypeError if there is no callback
   args // array, arguments
   // Returns: function
 ) => {
@@ -83,11 +83,10 @@ const requiredCallback = (
 };
 
 const onceCallback = (
-  // Extracts callback and make it safe
+  // Extract callback and make it safe
   // Wrap collback with once()
-  // and return common.emptiness if no callback
   args // array, arguments
-  // Returns: function
+  // Returns: function, callback or common.emptiness if there is no callback
 ) => {
   const callback = last(args);
   if (typeof callback === 'function') return once(args.pop());
--- a/lib/callbacks.js
+++ b/lib/callbacks.js
@@ -62,9 +62,9 @@ const unsafeCallback = (
 };
 
 const safeCallback = (
-  // Extract callback and return common.emptiness if there is no callback
+  // Extract callback
   args // array, arguments
-  // Returns: function, wrapped callback
+  // Returns: function, wrapped callback or common.emptiness if there is no one
   // Hint: previous name: `cbExtract` (deprecated)
 ) => {
   const callback = last(args);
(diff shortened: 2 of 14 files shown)


common.js
```diff
@@ -6,19 +6,19 @@ const submodules = [
...
   'network', // Network utilities
-  'id', // Kyes and identifiers
+  'id', // Keys and identifiers
   'sort', // Sort compare functions
   'cache', // Cache (enhanced Map)
   'mp', // Metaprogramming
   'enum', // Enumerated type
   'iterator', // Iterator
-].map(path => './lib/' + path).map(require);
+].map(path => require('./lib/' + path));
```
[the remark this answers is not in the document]
SemenchenkoVitaliy: Sorry, I didn't switched to master before creating a new branch. Is it better to just remove this commit or rebase onto master and start new PR?

lib/callbacks.js
```diff
@@ -83,11 +83,10 @@ const requiredCallback = (
 };
 
 const onceCallback = (
-  // Extracts callback and make it safe
+  // Extract callback and make it safe
   // Wrap collback with once()
```
belochub: s/collback/callback

lib/callbacks.js
```diff
@@ -62,7 +62,7 @@ const unsafeCallback = (
 };
 
 const safeCallback = (
-  // Extracts callback and return common.emptiness if no callback
+  // Extract callback and return common.emptiness if there is no callback
```
belochub: Should we maybe move part after 'return` to the `'Returns:'` description below?

lib/callbacks.js
```diff
@@ -73,7 +73,7 @@ const safeCallback = (
 };
 
 const requiredCallback = (
-  // Extracts callback and throw if no callback
+  // Extract callback and throw TypeError if there is no callback
```
belochub: Same question as above.

lib/data.js
```diff
@@ -60,7 +60,8 @@ function cloneObject(obj) {
 }
 
 const duplicate = (
-  // Duplicate object or array (properly handles prototype and circular links)
+  // Duplicate object or array
+  // It properly handles prototype and circular links
```
belochub: I think using `it` here is ambiguous, I'd prefer not to change that comment.

lib/math.js
```diff
@@ -26,13 +26,13 @@ class CryptoRandomPrefetcher {
 const cryptoPrefetcher = (
   // Create prefetcher to use when crypto.randomBytes is required to generate
   // multiple same-size values. `bufSize` must be a multiple of `valueSize` for
-  // this to work.
+  // this to work
```
belochub: Why did you remove a period here and leave it on line 53 below?
Also, I think it makes sense to have a period at the end of all full sentences when a comment has more than one sentence in it.

lib/mp.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
 const methods = (
-  // List method names
+  // Get list of method names
```
belochub: 'List' was a verb here, there's no need to change it.

lib/mp.js
```diff
@@ -16,7 +16,7 @@ const methods = (
 };
 
 const properties = (
-  // List property names
+  // Get list of property names
```
belochub: Same as above.

lib/strings.js
```diff
@@ -230,34 +231,37 @@ const arrayRegExp = (
 };
 
 const section = (
-  // Splits string by the first occurrence of separator
+  // Split string by the first occurrence of separator
   s, // string
   separator // string, or char
+  // Returns: array
```
belochub: Shouldn't it be 'array of strings' here?

lib/strings.js
```diff
@@ -230,34 +231,37 @@ const arrayRegExp = (
...
   if (i < 0) return [s, ''];
   return [s.slice(0, i), s.slice(i + separator.length)];
 };
 
 const rsection = (
-  // Splits string by the last occurrence of separator
+  // Split string by the last occurrence of separator
   s, // string
   separator // string, or char
+  // Returns: array
```
belochub: Same question as above.

lib/strings.js
```diff
@@ -230,34 +231,37 @@ const arrayRegExp = (
...
 };
 
 const split = (
-  // Splits string by multiple occurrence of separator
+  // Split string by multiple occurrence of separator
   s, // string
   separator = ',', // string (optional), default: ','
-  limit = -1 // number (optional), // max length of result array
+  limit = -1 // number (optional), max length of result array
+  // Returns: array
```
belochub: Same as above.

lib/strings.js
```diff
@@ -267,10 +271,11 @@ const split = (
 );
 
 const rsplit = (
-  // Splits string by multiple occurrence of separator
+  // Split string by multiple occurrences of separator
   s, // string
   separator = ',', // string (optional), default: ','
-  limit = -1 // number (optional), // max length of result array
+  limit = -1 // number (optional), max length of result array
+  // Returns: array
```
belochub: Ditto.

lib/units.js
```diff
@@ -81,7 +81,7 @@ const UNIT_SIZES = {
 };
 
 const sizeToBytes = (
-  // Convert string with units to int
+  // Convert string with file size to int
```
belochub: This is not necessarily a file size.

belochub: Also, s/int/integer, to be consistent with other places where you've changed it.

lib/units.js
```diff
@@ -57,7 +57,7 @@ const SIZE_UNITS = [
 ];
 
 const bytesToSize = (
-  // Convert int to string size Kb, Mb, Gb and Tb
+  // Convert int to string, representing file size in Kb, Mb, Gb and Tb
```
belochub: Same remarks as for the `sizeToBytes` function apply here.

belochub: Also, there should probably be a comma after 'Gb'.

tshemsedinov:
Landed in e75d985

Issue #183, opened 2018-09-01
Build for browsers


tshemsedinov:

tshemsedinov:
@nechaido I saw what you prepared babel build. @aqrln is it enough or we need webpack bundle?

aqrln:
@tshemsedinov it is enough, but we can update Babel to 7.x while we are there.

PR #184, opened 2018-09-03
Update Babel to 7.x.x

aqrln:
Fixes: https://github.com/metarhia/common/issues/183

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["env"]
+  "presets": ["@babel/preset-env"]
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": ["@babel/env"]
 }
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -1,3 +1,3 @@
 {
-  "presets": ["env"]
+  "presets": ["@babel/preset-env"]
```
belochub: You can shorten this to `@babel/env`.

.babelrc
```diff
@@ -1,3 +1,3 @@
 {
-  "presets": ["env"]
+  "presets": ["@babel/env"]
```
belochub: Should we also add `"sourceType": script` here similar to https://github.com/metarhia/jstp/pull/378?

aqrln: @belochub I don't think we should, since this option is undocumented and it won't change anything: we don't use webpack there, so we don't really care about CJS vs ESM in this case.

belochub:
Landed in https://github.com/metarhia/common/commit/c6a431774b9cb31755f52e44c9be48656dcb596a.

PR #186, opened 2018-09-08
Fix Iterator.some with thisArg test

lundibundi:
It used every function instead of some.

### DIFF
--- a/test/iterator.js
+++ b/test/iterator.js
@@ -283,13 +283,13 @@ metatests.test('Iterator.some that must return false', test => {
 
 metatests.test('Iterator.some with thisArg', test => {
   const obj = {
-    max: 0,
+    max: 2,
     predicate(value) {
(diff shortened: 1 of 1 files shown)


test/iterator.js
```diff
@@ -289,7 +289,7 @@ metatests.test('Iterator.some with thisArg', test => {
     },
   };
 
-  test.assertNot(iter(array).every(obj.predicate, obj));
+  test.assertNot(iter(array).some(obj.predicate, obj));
```
belochub: What do you think about setting the `max` value to `2` so that test would fail with `every` and pass only with `some`?

lundibundi: Ok.

belochub:
Landed in https://github.com/metarhia/common/commit/6289b26adab293eaf8695be227668f2597a3807b.

PR #188, opened 2018-09-08
Implement takeWhile iterator method

lundibundi:
It will return the values from the iterator as long as they satisfy the
predicate.

### DIFF
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -136,6 +136,10 @@ class Iterator {
     return new TakeIterator(this, amount);
   }
 
+  takeWhile(predicate, thisArg) {
+    return new TakeWhileIterator(this, predicate, thisArg);
+  }
+
   skip(amount) {
     for (let i = 0; i < amount; i++) {
       this.next();
@@ -255,6 +259,25 @@ class TakeIterator extends Iterator {
     if (this.iterated <= this.amount) {
       return this.base.next();
     }
+    return { done: true, value: undefined };
+  }
+}
+
+class TakeWhileIterator extends Iterator {
+  constructor(base, predicate, thisArg) {
+    super(base);
+    this.predicate = predicate;
+    this.thisArg = thisArg;
+    this.done = false;
+  }
+
+  next() {
+    if (this.done) return { done: true, value: undefined };
+    const next = this.base.next();
+    if (!next.done && this.predicate.call(this.thisArg, next.value)) {
+      return next;
+    }
+    this.done = true;
     return { done: true, value: undefined };
   }
 }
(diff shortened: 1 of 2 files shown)


lib/iterator.js
```diff
@@ -249,6 +253,28 @@ class TakeIterator extends Iterator {
...
+  constructor(base, predicate, thisArg) {
+    super(base);
+    this.predicate = predicate;
+    this.thisArg = thisArg;
+    this.done = false;
+  }
+
+  next() {
+    if (this.done) return { done: true, value: undefined };
+    for (const value of this.base) {
```
belochub: I don't think you need a loop here since you are always making only one iteration.

test/iterator.js
```diff
@@ -241,6 +241,12 @@ metatests.test('Iterator.take', test => {
   test.end();
 });
 
+metatests.testSync('Iterator.takeWhile', test => {
+  const it = iter(array).takeWhile(x => x < 3);
+  test.strictSame(it.toArray(), [1, 2]);
+  test.assert(it.next().done);
```
belochub: What is the point of this assert?
If `done` value is `false` here, then `toArray`, in the line before that, isn't working correctly, so you are testing the `toArray` method with this assert.

lundibundi: I test that after we exited once iterator won't continue again.
This can be easily simulated with returning `done: false` in the loop but not setting `this.done` variable and having array of sort [1, 2, 3, 1, 1] for this case.

lib/iterator.js
lundibundi: I don't think I understand, I iterate over the base iterator, therefore, this loop will iterate over all of base-iterator values untill `predicate` return false.

test/iterator.js
belochub: Okay, I'm fine with it, but that still doesn't guarantee that if you call `next()` method one more time it won't continue again, meaning that to test it fully we need to add an infinite amount of checks for the `next()` result, which is impossible to do.

lib/iterator.js
belochub: You only have an `if` statement inside this loop and in each of the `if` statement's branches you exit the loop, meaning that you always make only one iteration.

test/iterator.js
lundibundi: Yeah, but at least we check that we didn't make an unintentional error of missing to check/set `done`. Obviously, this won't protect against a manually crafted implementation that will avoid this check.

lib/iterator.js
lundibundi: Oh, good catch. I just copy-pasted Filter impl here :upside_down_face:.

belochub:
Landed in https://github.com/metarhia/common/commit/dffbb3343e95e95a1837280708c17a6d4c6bf20e.

PR #191, opened 2018-09-16
Fix duplicate

tshemsedinov:


tshemsedinov:
Landed in b2f7cf2 and 61c733b

PR #194, opened 2018-09-21
Change comments style

SemenchenkoVitaliy:


lib/array.js
```diff
@@ -35,17 +32,16 @@ const range = (
...
-  // list: sequence([81, 82, 83]) = [81, 82, 83]
-  // range from..to: sequence([81,,83]) = [81, 82, 83]
-  // range from..count: sequence([81, [3]]) = [81, 82, 83]
-  // range from..max-to: sequence([81, [-2]], 5) = [81, 82, 83]
-) => {
+// Generate int array from sequence syntax
+//   seq - array
+//   max - number, optional max
+// Returns: array
+// Example:
```
lundibundi: Should we change this to 'Examples:' everywhere?
OR
We might use 'Example:' prefix for each example in order to cleanly separate them and parse/display properly later on.

I personally prefer the latter style, as it is more descriptive, understandable for the reader and supports multiline examples simply.
@SemenchenkoVitaliy @belochub @nechaido @tshemsedinov wdyt?

lib/cache.js
```diff
@@ -13,11 +13,10 @@ util.inherits(Cache, Map);
...
 const dataSize = data => (data && data.length ? data.length : 0);
 
-Cache.prototype.add = function(
-  // Add key-value pair to cache
-  key, // string, key
-  val // value, vaule
-) {
+// Add key-value pair to cache
+//   key - string, key
+//   val - value, vaule
```
lundibundi: Typo in 'value'.
Also, type should be 'any', but I'm not sure, maybe we have specific name for it?

lib/cache.js
```diff
@@ -26,22 +25,20 @@ Cache.prototype.add = function(
...
 };
 
-Cache.prototype.clr = function(
-  // Clear cache elements that start with prefix
-  prefix, // string, to compare with beginning of the key
-  fn // function (optional), `(key, val)` to be called on each key
-) {
+// Clear cache elements that start with prefix
+//   prefix - string, to compare with beginning of the key
+//   fn - function (optional), `(key, val)` to be called on each key
```
lundibundi: Strange ordering, shouldn't function signature be next to the function keyword as it's actually a type?

lib/cache.js
```diff
@@ -51,14 +48,13 @@ Cache.prototype.clr = function(
...
 
-const cache = (
-  // Extend Map interface with:
-  // `cache.allocated` - total allocated size
-  // `cache.add(key, val)` - add record
-  // `cache.del(key)` - delete record
-  // `cache.clr(prefix, fn)` - delete all if `key.startsWith(prefix)`
-  // Returns: object, cache instance
-) => (new Cache());
+// Extend Map interface with:
```
lundibundi: We can probably replace this whole comment with just
Returns new Cache instance
Or something? As all of the explanation will be present in the Cache class documentation.

lib/callbacks.js
```diff
@@ -48,46 +40,42 @@ const once = (
...
 
-const safeCallback = (
-  // Extract callback
-  args // array, arguments
-  // Returns: function, wrapped callback or common.emptiness if there is no one
-  // Hint: previous name: `cbExtract` (deprecated)
-) => {
+// Extract callback
+//   args - array, arguments
+// Returns: function, wrapped callback or common.emptiness if there is no one
```
lundibundi: Call back is not wrapped in any way here, should probably be just ' callback or comm....'.

lib/callbacks.js
```diff
@@ -48,46 +40,42 @@ const once = (
...
 };
 
-const requiredCallback = (
-  // Extract callback
-  args // array, arguments
-  // Returns: function or throw TypeError if there is no callback
-) => {
+// Extract callback
+//   args - array, arguments
+// Returns: function or throw TypeError if there is no callback
```
lundibundi: Should we consider adding 'Throws' clause for such cases?
/cc @SemenchenkoVitaliy @belochub @nechaido @tshemsedinov wdyt?

lib/callbacks.js
```diff
@@ -48,46 +40,42 @@ const once = (
...
-  // Returns: function, callback or null
-  // Hint: previous name: `common.cbUnsafe` (deprecated)
-  // Hint: another alias: `common.extractCallback` (deprecated)
-) => {
+// Extract callback function
+// It's unsafe: may return null, allows multiple calls
+//   args - array, arguments
+// Returns: function, callback or null
+// Hint: previous name: `common.cbUnsafe` (deprecated)
+// Hint: another alias: `common.extractCallback` (deprecated)
```
lundibundi: Also, add some sort of 'deprecated' label?
/cc @SemenchenkoVitaliy @belochub @nechaido @tshemsedinov wdyt?

lib/fp.js
```diff
@@ -2,32 +2,29 @@
...
-const omap = (
-  // Map object fields with provided function
-  mapFn, // funtion, to appy to every field value
-  obj // object, which fields used for mapping
-  // Returns: object, with same reference but with transformed fields
-) => {
+// Partially apply arguments to function
+//   fn - function
+//   args - array, argumants
+// Returns: function
```
lundibundi: This should probably be `function(...args), args - arguments`  (Also applies in other places)

lib/fp.js
```diff
@@ -2,32 +2,29 @@
...
 
-const compose = (
-  // Compose multiple functions into one
-  ...fns // array of functions
-  // Returns: function, composed
-) => (
-  ...args // arguments to be passed to first function
-) => {
+// Compose multiple functions into one
+//   fns - array of functions
```
lundibundi: This is not an array, you just pass functions here.

lib/fp.js
```diff
@@ -2,32 +2,29 @@
...
+// Returns: function, composed
```
lundibundi: Also, `function(...args) ......`  (Also applies in other places)

lib/fp.js
```diff
@@ -37,23 +34,21 @@ const compose = (
...
-  // Zip several arrays into one
-  ...arrays // array of arrays,
-  // Returns: array, length is minimal of input arrays length
-  // Element with index i of resulting array is array with
-  // elements with index i from input arrays
-) => {
+// Zip several arrays into one
+//   arrays - array of arrays,
+// Returns: array, length is minimal of input arrays length
+// Element with index i of resulting array is array with
```
lundibundi: The description should be at the top. (Also applies in other places)

lib/fp.js
```diff
@@ -71,31 +66,28 @@ const zip = (
...
+//   count - number, new array length
+//   elem - value to replicate
+// Returns: array, replicated
+const replicate = (count, elem) => (Array.from({ length: count }, () => elem));
+
+// Zip arrays using specific function
+//   fn - function, for zipping elements with index i
+//   arrays - array of arrays
+// Returns: array
+// Element with index i of resulting array is result
```
lundibundi: Ditto.

lib/fp.js
```diff
@@ -71,31 +66,28 @@ const zip = (
...
+// Create array of replicated values
+//   count - number, new array length
+//   elem - value to replicate
+// Returns: array, replicated
+const replicate = (count, elem) => (Array.from({ length: count }, () => elem));
+
+// Zip arrays using specific function
+//   fn - function, for zipping elements with index i
+//   arrays - array of arrays
+// Returns: array
```
lundibundi: Elaborate on the meaning of array (data in it) here please. (Also applies in other places)

lib/callbacks.js
tshemsedinov: Don't think so. Now we need quick decisions and results, have no time to prepare whole special language for commenting.

lib/fp.js
tshemsedinov: `fns` becomes an array after rest operator, so I think we will describe identifier type

lib/cache.js
lundibundi: I think we can just use 
`// fn - function(key, val), to be called on each key`
?

lib/fp.js
lundibundi: IMO those descriptions are not code-descriptions, they are usage-descriptions, therefore, they must be written from the point of view of a caller.

Hence here if a caller doesn't know the signature and just read comments/generated docs he may (and should) assume that he must pass an array to this function but (`compose([fn1, fn2,fn3])`) will lead to an incorrect behavior.

I think the best option will be to have it as follows (include rest and description in a doc)
`//    ...fns - functions to be composed`
OR
`//    ...fns - array, functions to be composed`

lib/cache.js
SemenchenkoVitaliy: Well, to be parsed correctly we have to have something like this
```
//   fn - function, function(key, val), to be called on each key
```
or
```
//   fn - function (optional), to be called on each key
//     key - string, key
//     val - value
```

lib/fp.js
SemenchenkoVitaliy: As for me it belongs more to `Returns` than to function description

lib/cache.js
lundibundi: I'd prefer the second variant but add the signature in the first type
`fn - function(key, val) (optional)`
Also note that the comment to the function might be multiline.

lib/callbacks.js
lundibundi: It's the same syntax as `Deprecate` + `Argument` so I don't think it will be complicated.
`Throws: Error, when there is no callback`
`Throws: type, description`

lib/callbacks.js
SemenchenkoVitaliy: I'm ok with that, but in this case it's just renaming issue and since this particular function is used there is no need for deprecation label. However, I'll consider using another label for such cases. The ones can be `Alias` or `Previous name`. WDYT?

lib/fp.js
lundibundi: Oh, I thought it's not about the return value. Could you then fix the grammar of this sentence / add punctuation to make it a whole sentence please.

lib/fp.js
lundibundi: To clarify, there must not be such linebreak between return type and it's description. also, 'is result' at the end there makes no sense.

SemenchenkoVitaliy: It's part of sentence `element with index i of resulting array is result of fn called with arguments from arrays`

lib/cache.js
```diff
@@ -15,7 +15,7 @@ const dataSize = data => (data && data.length ? data.length : 0);
 
 // Add key-value pair to cache
 //   key - string, key
-//   val - value, vaule
+//   val - any type, associative value
```
lundibundi: s/associative/associated

lib/cache.js
```diff
@@ -37,7 +37,9 @@ Cache.prototype.del = function(key) {
 
 // Clear cache elements that start with prefix
 //   prefix - string, to compare with beginning of the key
-//   fn - function (optional), `(key, val)` to be called on each key
+//   fn - function, (key, val) (optional)
```
lundibundi: Can we make this without comma? (i.e. `fn - function(key, val) (optional)`)

lib/cache.js
```diff
@@ -48,12 +50,8 @@ Cache.prototype.clr = function(prefix, fn) {
...
 };
 
-// Extend Map interface with:
-// `cache.allocated` - total allocated size
-// `cache.add(key, val)` - add record
-// `cache.del(key)` - delete record
-// `cache.clr(prefix, fn)` - delete all if `key.startsWith(prefix)`
-// Returns: object, cache instance
+// Get Cache, enhanced Map
+// Returns: Cache, instance
```
lundibundi: Unneded comma.

lib/cache.js
```diff
@@ -48,12 +50,8 @@ Cache.prototype.clr = function(prefix, fn) {
   });
 };
 
-// Extend Map interface with:
-// `cache.allocated` - total allocated size
-// `cache.add(key, val)` - add record
-// `cache.del(key)` - delete record
-// `cache.clr(prefix, fn)` - delete all if `key.startsWith(prefix)`
-// Returns: object, cache instance
+// Get Cache, enhanced Map
```
lundibundi: s/Get Cache, enhanced Map/Create new Cache instance/

lib/callbacks.js
```diff
@@ -44,8 +44,7 @@ const once = fn => {
 // It's unsafe: may return null, allows multiple calls
 //   args - array, arguments
 // Returns: function, callback or null
-// Hint: previous name: `common.cbUnsafe` (deprecated)
-// Hint: another alias: `common.extractCallback` (deprecated)
+// Deprecated: previous names: `common.cbUnsafe`, `common.extractCallback`
```
lundibundi: Is `unsafeCallback` deprecated? Because this comment tells that it is.
I think that this comment meant that the previous name (common.cbUnsafe) is deprecated, not this function.

lib/callbacks.js
```diff
@@ -54,8 +53,8 @@ const unsafeCallback = args => {
 
 // Extract callback
 //   args - array, arguments
-// Returns: function, wrapped callback or common.emptiness if there is no one
-// Hint: previous name: `cbExtract` (deprecated)
+// Returns: function, callback or common.emptiness if there is no one
+// Deprecated: previous name: `common.cbExtract`
```
lundibundi: Ditto.

lib/fp.js
```diff
@@ -4,9 +4,9 @@ const { safeCallback } = require('./callbacks');
 
 // Partially apply arguments to function
 //   fn - function
-//   args - array, argumants
-// Returns: function
-//   rest - arguments
+//   ...args - array, argumants to be applied
```
lundibundi: s/argumants/arguments

lib/fp.js
```diff
@@ -74,19 +74,20 @@ const replicate = (count, elem) => (Array.from({ length: count }, () => elem));
 
 // Zip arrays using specific function
 //   fn - function, for zipping elements with index i
-//   arrays - array of arrays
-// Returns: array
-// Element with index i of resulting array is result
+//   ...arrays - array, arrays to be zipped
+// Returns: array, zipped
+// Hint: Element with index i of resulting array is result
 // of fn called with arguments from arrays
```
lundibundi: I don't think we should use `Hint` keyword, we can just put everything in the place it belongs to (here Result).

Why not 
```
// Returns: array, contains element zipped with provided function. Element at index i of the
//          resulting array is the result of calling fn with arguments at the ith position
//          from input arrays.
```
?
Similar can be applied in other places.

lib/cache.js
SemenchenkoVitaliy: This can be done, but for this purpose parser need to be changed to support explicitly this style of commenting functions arguments. One option is to change it as you suggested and leave it until parser will be ready to support such comments type.

lib/callbacks.js
SemenchenkoVitaliy: That's why is asked it here in https://github.com/metarhia/common/pull/194#discussion_r219679795. We need special label to store previous names. Other options are to remove them at all or copy-paste code from the time when they were present, but I'd like not to have something like this

lib/cache.js
SemenchenkoVitaliy: There are similar comments in `events` https://github.com/metarhia/common/blob/comments-refactor/lib/events.js#L68 Should they be also changed?

lib/time.js
```diff
@@ -1,22 +1,20 @@
...
 
 const pad2 = n => (n < 10 ? '0' + n : '' + n);
 
-const nowDate = (
-  // Get current date in YYYY-MM-DD format
-  date // Date (optional), now by default
-  // Returns: string
-) => {
+// Get current date in YYYY-MM-DD format
+//   date - Date (optional), default: Date.now()
```
belochub: The default value for this argument is not `Date.now()`, but `new Date()`.

lib/time.js
```diff
@@ -25,11 +23,10 @@ const nowDate = (
   );
 };
 
-const nowDateTime = (
-  // Get current date in YYYY-MM-DD hh:mm format
-  date // Date (optional), now by default
-  // Returns: string
-) => {
+// Get current date in YYYY-MM-DD hh:mm format
+//   date - Date (optional), default: Date.now()
```
belochub: Ditto.

lib/utilities.js
```diff
@@ -4,11 +4,11 @@ const { basename } = require('path');
 
 const { between } = require('./strings');
 
-const deprecate = (
-  // Wrap method to mark it as deprecated
-  fn // function (optional)
-  // Returns: function, wrapped with deprecation warning
-) => {
+// Wrap method to mark it as deprecated
+//   fn - function (optional)
```
belochub: Is `fn` really optional here?

lib/utilities.js
```diff
@@ -21,11 +21,11 @@ const deprecate = (
   return wrap;
 };
 
-const alias = (
-  // Wrap new method to mark old alias as deprecated
-  fn // function (optional)
-  // Returns: function, wrapped with deprecation warning
-) => {
+// Wrap new method to mark old alias as deprecated
+//   fn - function (optional)
```
belochub: Same question as above.

lib/utilities.js
```diff
@@ -21,11 +21,11 @@ const deprecate = (
...
+// Returns: function,  function(...args), wrapped with deprecation warning
+//   ...args - array, arguments to be passed to wrapped funtion
```
belochub: Typo in word 'funtion'.

lib/utilities.js
```diff
@@ -4,11 +4,11 @@ const { basename } = require('path');
...
 
-const deprecate = (
-  // Wrap method to mark it as deprecated
-  fn // function (optional)
-  // Returns: function, wrapped with deprecation warning
-) => {
+// Wrap method to mark it as deprecated
+//   fn - function (optional)
+// Returns: function, (...args), wrapped with deprecation warning
+//   ...args - array, arguments to be passed to wrapped funtion
```
belochub: Typo in word 'funtion'.

lib/utilities.js
```diff
@@ -41,11 +41,11 @@ const alias = (
...
 
-const safe = (
-  // Make function raise-safe
-  fn // function
-  // Returns: function, wrapped with try/catch interception
-) => (...args) => {
+// Make function raise-safe
+//   fn - function
+// Returns: function, wrapped with try/catch interception
+//   ...args - array, arguments to be passed to wrapped funtion
```
belochub: Typo in word 'funtion'.

lib/cache.js
lundibundi: But you are already changing the parser in https://github.com/metarhia/metaschema/pull/107, so is there any problem of including this syntax there? The current parser cannot parse any of the styles here.

lib/cache.js
lundibundi: Yeah, but be sure to add the description to the `EnhancedEmmiter` class.

lib/callbacks.js
lundibundi: We don't care about previous names, that's not how deprecation works. You deprecate the *previous* function with smth like `Deprecated: use newName instead` and the `newName` don't have anything about the previous names.

If the `common.cbUnsafe`, `common.extractCallback` are already removed then just remove this comment as it is not needed.

lib/events.js
```diff
@@ -23,7 +23,7 @@ const forwardEvent = (from, to, event, newEvent = event) => {
 // Forward events from one EventEmitter to another
 //   from - EventEmitter, to listen for event
 //   to - EventEmitter, to emit event on
-//   events - array of strings, events names
+//   events - array of strings, string or object, events names
```
belochub: It is also optional since you can omit it, according to examples.

lib/id.js
```diff
@@ -50,7 +50,7 @@ const generateGUID = () => {
 };
 
 // Calculate SID CRC
-//   config - record, { length, characters, secret }
+//   config - record, { secret }
```
belochub: You haven't addressed the second part of the https://github.com/metarhia/common/pull/194#discussion_r219775382:
>Also, the first word after the dash describes the type in all of the other comments, and 'record' here should probably be replaced with an 'object' since this function is expecting an object.

lib/id.js
```diff
@@ -73,7 +73,7 @@ const generateSID = config => {
 };
 
 // Validate SID
-//   config - record, { length, characters, secret }
+//   config - record, { secret }
```
belochub: Ditto.

lundibundi:
Based on this [`array of strings, string or object`](https://github.com/metarhia/common/pull/194/files#diff-71dcd327d0ca067b490b22d677f81966R26) should we add some special syntax for such cases, like
`array<string> | string | object` to separate it from other comments (`optional`, text etc) and not use commas everywhere?

SemenchenkoVitaliy:
> Based on this array of strings, string or object should we add some special syntax for such cases, like
array<string> | string | object to separate it from other comments (optional, text etc) and not use commas everywhere?

I like this idea, however, it's not possible if we have to maintain backward compatibility.

lundibundi:
I don't think there is anything more to break =). If you are referring to https://github.com/metarhia/metaschema/pull/107#issuecomment-423511723 then it never supported multiple types (i.e. `string or object`) therefore the behavior will be the same from the point of view of `guard`(https://github.com/metarhia/metaschema/blob/master/lib/schema.js#L289).

SemenchenkoVitaliy:
This requires changes in `createInstance` which can be done, but I'd like to hear opinions, because now only `string`, `number`, `boolean`, `undefined`, `function`, `object`, `null`, `symbol` are supported. Also there are some functions which use capital-case `Object` and `scalar`.
BTW, I think this conversation should be moved to the appropriate [pull request](https://github.com/metarhia/metaschema/pull/107).

lib/utilities.js
```diff
@@ -41,11 +41,11 @@ const alias = (
...
 
-const safe = (
-  // Make function raise-safe
-  fn // function
-  // Returns: function, wrapped with try/catch interception
-) => (...args) => {
+// Make function raise-safe
+//   fn - function
+// Returns: function, wrapped with try/catch interception
+//   ...args - array, arguments to be passed to wrapped function
```
tshemsedinov: I think descriding arguments we do not need `...`
So not `...args - array` but `args - array`

lib/strings.js
```diff
@@ -230,55 +216,51 @@ const arrayRegExp = (
...
-  // Split string by the first occurrence of separator
-  s, // string
-  separator // string, or char
-  // Returns: array of strings
-  // Example: rsection('All you need is JavaScript', 'is')
-  // Result: ['All you need ', ' JavaScript']
-) => {
+// Split string by the first occurrence of separator
+//   s - string
+//   separator - string, or char
```
tshemsedinov: `string or char`

lib/fp.js
```diff
@@ -166,11 +154,11 @@ const either = (
...
 
-const restLeft = (
-  // Rest left, transform function
-  fn // function, (args, arg1..argN, callback)
-  // Returns: function, (arg1..argN, ...args, callback)
-) => (...spreadArgs) => {
+// Rest left, transform function
+//   fn - function(args, arg1..argN, callback)
+// Returns: function(arg1..argN, ...args, callback)
+//   ...spreadArgs - array, arguments to be added
```
tshemsedinov:
```
arg1..argN - description
args - description
callback - description
```

lib/utilities.js
tshemsedinov: Fix this in all places

SemenchenkoVitaliy:
@tshemsedinov I'd like to hear your opinion about https://github.com/metarhia/common/pull/194#issuecomment-423983573. @lundibundi wrote:
> Based on this [`array of strings, string or object`](https://github.com/metarhia/common/pull/194/files#diff-71dcd327d0ca067b490b22d677f81966R26) should we add some special syntax for such cases, like
`array<string> | string | object` to separate it from other comments (`optional`, text etc) and not use commas everywhere?

Also, which theese long comments style is preferable?
```
//   arg - long description,
//     rest of long description
```
or
```
//   arg - long description,
//   rest of long description
```

lundibundi:
The current style for multiline arg-comments is
```
//   arg - long description,
//         rest of long description
```
(aligned with the top line)

lib/events.js
```diff
@@ -21,16 +20,15 @@ const forwardEvent = (
...
-  // Example: common.forwardEvent(from, to);
-  // Example: common.forwardEvent(from, to, 'eventName');
-  // Example: common.forwardEvent(from, to, { eventName: 'newEventName' });
-  // Example: common.forwardEvent(from, to, ['eventName1', 'eventName2']);
-) => {
+// Forward events from one EventEmitter to another
+//   from - EventEmitter, to listen for event
+//   to - EventEmitter, to emit event on
+//   events - array of strings, string or object, (optional), events names
+// Example: common.forwardEvent(from, to);
```
nechaido: This should be `forwardEvents`, look like a typo in an original code. We might as well fix it in this PR but in a separate commit.

lib/utilities.js
SemenchenkoVitaliy: Furthermore, when generating docs, we will get
```md
### Function description
`fn(args)`
```
instead of
```md
### Function description
`fn(...args)`
```

burbokop:
https://github.com/metarhia/common/pull/202

SemenchenkoVitaliy:
@burbokop There is no need to create another similar PR in this repository. Moreover, your changes were already fixed [here](https://github.com/metarhia/common/pull/194/files#diff-63b6fc24d5d394b84bf4b09d46dc07e5R39).

lib/events.js
```diff
@@ -21,16 +20,15 @@ const forwardEvent = (
...
-  // Example: common.forwardEvent(from, to);
-  // Example: common.forwardEvent(from, to, 'eventName');
-  // Example: common.forwardEvent(from, to, { eventName: 'newEventName' });
-  // Example: common.forwardEvent(from, to, ['eventName1', 'eventName2']);
-) => {
+// Forward events from one EventEmitter to another
+//   from - EventEmitter, to listen for event
+//   to - EventEmitter, to emit event on
+//   events - <string[]>, string or object, (optional), events names
+// Example: common.forwardEvents(from, to);
```
lundibundi: I don't think we need to have `common` in these (and other such examples).

lib/events.js
```diff
@@ -58,19 +56,24 @@ util.inherits(EnhancedEmitter, EventEmitter);
...
+// Call listener with provided arguments
+//   args - <array>, arguments to be passed
 EnhancedEmitter.prototype.emit = function(...args) {
   emit.call(this, '*', ...args);
   emit.call(this, ...args);
 };
 
+// Forward events from one EventEmitter to another
+//   to - EventEmitter, to emit event on
+//   events - <string[]>, string or object, events names
```
lundibundi: Shouldn't this be
`<string[]> | <string> | <object>`
?

lib/fp.js
```diff
@@ -2,32 +2,29 @@
...
-  // Map object fields with provided function
-  mapFn, // funtion, to appy to every field value
-  obj // object, which fields used for mapping
-  // Returns: object, with same reference but with transformed fields
-) => {
+// Partially apply arguments to function
+//   fn - <function>
+//   args - <array>, arguments to be applied
+// Returns: function(...rest)
+//   rest - <array>, arguments
```
lundibundi: I still don't think `<array>` conveys the meaning of `rest` here.
What if we have special case like we have for an array (`<string[]>`) but instead it will be `<string...>`, wdyt?

lib/fp.js
```diff
@@ -37,23 +34,21 @@ const compose = (
...
-  // Apply given function to value or default value
-  fn, // function
-  defVal, // default value
-  value // value (optional)
-  // Returns: result of `fn` or `defVal`
-) => (
+// Apply given function to value or default value
+//   fn - <function>
+//   defVal - default value
+//   value - value (optional)
```
lundibundi: I propose to use square brackets around the argument name to convey 'optionality' of the argument. i.e. `[value] - <any>, value`?

Also this doesn't have types for values.

lib/fp.js
lundibundi: Could you please align the `Returns` cases?
i.e
```
Returns: <array>, zipped, element with index i of resulting array is result
         of fn called with arguments from arrays.
```

Also, yeah, doesn't have types in <> for Returns case.

lib/sort.js
```diff
@@ -1,12 +1,11 @@
...
-  priority, // array of strings with priority
-  s1, s2 // string, to compare
-  // Returns: number
-  // Example: files.sort(common.sortComparePriority)
-) => {
+// Compare for array.sort with priority
+//   priority - <string[]>, with priority
+//   s1, s2 - <string>, to compare
+// Returns: <number>
+// Example: files.sort(common.sortComparePriority)
```
lundibundi: I'd suggest to always add blank line between 'function description, args, returns, throws' and 'examples'.

lib/strings.js
```diff
@@ -88,45 +86,41 @@ const subst = (
...
-  fileName // string, file name
-  // Returns: string
-  // Example: fileExt('/dir/file.txt')
-  // Result: 'txt'
-) => (
+// Extract file extension in lower case with no dot
+//   fileName - <string>, file name
+// Returns: <string>
+// Example: fileExt('/dir/file.txt')
+// Result: 'txt'
```
lundibundi: 'txt'?
maybe `Result: <string>, file extension without dot`?

lib/strings.js
```diff
@@ -88,45 +86,41 @@ const subst = (
...
 };
 
-const fileExt = (
-  // Extract file extension in lower case with no dot
-  fileName // string, file name
-  // Returns: string
-  // Example: fileExt('/dir/file.txt')
-  // Result: 'txt'
-) => (
+// Extract file extension in lower case with no dot
```
lundibundi:
```suggestion
// Extract file extension in lower case without dot
```

lib/strings.js
```diff
@@ -88,45 +86,41 @@ const subst = (
...
-  fileName // string, file name
-  // Returns: string
-  // Example: fileExt('file.txt')
-  // Result: 'file'
-) => (
+// Remove file extension from file name
+//   fileName - <string>, file name
+// Returns: <string>
+// Example: fileExt('file.txt')
+// Result: 'file'
```
lundibundi: Same as above.

lib/strings.js
```diff
@@ -230,55 +216,51 @@ const arrayRegExp = (
...
-  // Returns: array of strings
-  // Example: rsection('All you need is JavaScript', 'is')
-  // Result: ['All you need ', ' JavaScript']
-) => {
+// Split string by the first occurrence of separator
+//   s - <string>
+//   separator - <string>, or char
+// Returns: <string[]>
+// Example: rsection('All you need is JavaScript', 'is')
+// Result: ['All you need ', ' JavaScript']
```
lundibundi: I don't consider this a good idea to have special `Result` for `Example`, perhaps we could put it this way:
```
// Example: rsection('All you need is JavaScript', 'is')
//          -> ['All you need ', ' JavaScript']
```
OR
```
// Example: rsection('All you need is JavaScript', 'is') -> ['All you need ', ' JavaScript']
```
?

lib/strings.js
belochub: @lundibundi, `'txt'` here is the result of running the example, the text you are proposing must be above the example in 'Returns:', describing what to expect in general as a result of calling this function.

lib/time.js
```diff
@@ -1,22 +1,20 @@
...
 
 const pad2 = n => (n < 10 ? '0' + n : '' + n);
 
-const nowDate = (
-  // Get current date in YYYY-MM-DD format
-  date // Date (optional), now by default
-  // Returns: string
-) => {
+// Get current date in YYYY-MM-DD format
+//   date - Date (optional), default: new Date()
```
belochub:
```suggestion
//   date - <Date> (optional), default: new Date()
```

lib/time.js
```diff
@@ -1,22 +1,20 @@
...
-  // Compare time1 and time2
-  time1, // string, time or milliseconds
-  time2 // string, time or milliseconds
-  // Returns: boolean
-  // Example: isTimeEqual(sinceTime, buffer.stats.mtime)
-) => (
+// Compare time1 and time2
+//   time1 - <string>, time or milliseconds
+//   time2 - <string>, time or milliseconds
+// Returns: boolean
```
belochub:
```suggestion
// Returns: <boolean>
```

lib/time.js
```diff
@@ -1,22 +1,20 @@
...
 const pad2 = n => (n < 10 ? '0' + n : '' + n);
 
-const nowDate = (
-  // Get current date in YYYY-MM-DD format
-  date // Date (optional), now by default
-  // Returns: string
-) => {
+// Get current date in YYYY-MM-DD format
+//   date - Date (optional), default: new Date()
+// Returns: string
```
belochub:
```suggestion
// Returns: <string>
```

lib/utilities.js
```diff
@@ -41,11 +41,11 @@ const alias = (
...
 };
 
-const safe = (
-  // Make function raise-safe
-  fn // function
-  // Returns: function, wrapped with try/catch interception
-) => (...args) => {
+// Make function raise-safe
+//   fn - <function>
+// Returns: function, wrapped with try/catch interception
```
belochub:
```suggestion
// Returns: <function>, wrapped with try/catch interception
```

lib/utilities.js
```diff
@@ -21,11 +21,11 @@ const deprecate = (
...
 };
 
-const alias = (
-  // Wrap new method to mark old alias as deprecated
-  fn // function (optional)
-  // Returns: function, wrapped with deprecation warning
-) => {
+// Wrap new method to mark old alias as deprecated
+//   fn - <function>
+// Returns: function,  function(...args), wrapped with deprecation warning
```
belochub:
```suggestion
// Returns: <function>, function(...args), wrapped with deprecation warning
```

lib/strings.js
lundibundi: Yeah, I have made a note about this syntax but didn't update the comment.
https://github.com/metarhia/common/pull/194#discussion_r226683471

lib/callbacks.js
```diff
@@ -48,46 +40,40 @@ const once = (
   return wrap;
 };
 
-const unsafeCallback = (
-  // Extract callback function
-  // It's unsafe: may return null, allows multiple calls
-  args // array, arguments
-  // Returns: function, callback or null
-  // Hint: previous name: `common.cbUnsafe` (deprecated)
```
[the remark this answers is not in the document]
lundibundi: It'd make sense to have them if the function in question actually still existed (we would then make it an alias and write `Deprecated: renamed to..` notice there), but I don't think it's worth having such notes. Because realistically speaking they don't have a limit as they reference already deleted content and will only be arguably useful while cluttering the codebase and docs.

Though I agree that they could be useful but not in the form of code comment. Perhaps if we had a whole changelog for every version and therefore specific separate history for functions names based on that...

lib/fp.js
SemenchenkoVitaliy: But what about something like this: https://github.com/metarhia/common/blob/comments-refactor/lib/data.js#L180? Will we have `<Array[]>`, `<Array...>` or `<Array[]...>`?

lib/time.js
```diff
@@ -25,11 +24,10 @@ const nowDate = (
   );
 };
 
-const nowDateTime = (
-  // Get current date in YYYY-MM-DD hh:mm format
-  date // Date (optional), now by default
-  // Returns: string
-) => {
+// Get current date in YYYY-MM-DD hh:mm format
+//   date - Date, (optional), default: new Date()
```
belochub:
```suggestion
//   date - <Date>, (optional), default: new Date()
```

lib/utilities.js
```diff
@@ -4,11 +4,11 @@ const { basename } = require('path');
 
 const { between } = require('./strings');
 
-const deprecate = (
-  // Wrap method to mark it as deprecated
-  fn // function (optional)
-  // Returns: function, wrapped with deprecation warning
-) => {
+// Wrap method to mark it as deprecated
+//   fn - <function>
```
belochub: I propose to change it to `Function` here and everywhere else (similar to how it's done in Node.js docs):
```suggestion
//   fn - <Function>
```

lib/cache.js
```diff
@@ -26,22 +25,22 @@ Cache.prototype.add = function(
...
 };
 
-Cache.prototype.clr = function(
-  // Clear cache elements that start with prefix
-  prefix, // string, to compare with beginning of the key
-  fn // function (optional), `(key, val)` to be called on each key
-) {
+// Clear cache elements that start with prefix
+//   prefix - <string>, to compare with beginning of the key
+//   fn - <function>, fn(key, val) (optional)
```
belochub: I propose to change it to `Function` here and everywhere else (similar to how it's done in Node.js docs):
```suggestion
//   fn - <Function>, fn(key, val) (optional)
```

lib/cache.js
```diff
@@ -51,14 +50,9 @@ Cache.prototype.clr = function(
...
-const cache = (
-  // Extend Map interface with:
-  // `cache.allocated` - total allocated size
-  // `cache.add(key, val)` - add record
-  // `cache.del(key)` - delete record
-  // `cache.clr(prefix, fn)` - delete all if `key.startsWith(prefix)`
-  // Returns: object, cache instance
-) => (new Cache());
+// Create Cache, enhanced Map
+// Returns: <object>, Cache instance
```
belochub: Same here, I think `Object` will be better:
```suggestion
// Returns: <Object>, Cache instance
```

lib/callbacks.js
```diff
@@ -3,41 +3,33 @@
...
+// Empty function
+// Returns: <boolean>, always false
+const falseness = () => false;
+
+// Empty function
+// Returns: <boolean>, always true
+const trueness = () => true;
+
+// Empty function
+// Returns: <undefined>
```
belochub: I am not sure if this should be here since this function doesn't explicitly return `undefined`, it just doesn't return anything.

lib/utilities.js
SemenchenkoVitaliy: I like this idea, though it will require to change [types](https://github.com/metarhia/metaschema/blob/master/lib/types.js) in metaschema in order to properly parse parameters types. Another option is to to convert them to lowercase, but this doesn't look great to me.

lib/callbacks.js
SemenchenkoVitaliy: Return value was explicitly stated previously https://github.com/metarhia/common/blob/master/lib/callbacks.js#L18. And since it is supposed to return `undefined`, I think it'll be better to either leave it as it is or add `return undefined` to it.

lib/cache.js
```diff
@@ -26,22 +25,22 @@ Cache.prototype.add = function(
...
 };
 
-Cache.prototype.clr = function(
-  // Clear cache elements that start with prefix
-  prefix, // string, to compare with beginning of the key
-  fn // function (optional), `(key, val)` to be called on each key
-) {
+// Clear cache elements that start with prefix
+//   prefix - <string>, to compare with beginning of the key
+//   fn - <Function>, fn(key, val) (optional)
```
lundibundi: `fn(key, val)` can be generated from the arguments below and should probably be omitted here.

lib/cache.js
```diff
@@ -51,14 +50,9 @@ Cache.prototype.clr = function(
...
-const cache = (
-  // Extend Map interface with:
-  // `cache.allocated` - total allocated size
-  // `cache.add(key, val)` - add record
-  // `cache.del(key)` - delete record
-  // `cache.clr(prefix, fn)` - delete all if `key.startsWith(prefix)`
-  // Returns: object, cache instance
-) => (new Cache());
+// Create Cache, enhanced Map
+// Returns: <Object>, Cache instance
```
lundibundi:
```suggestion
// Returns: <Cache>
```

lib/callbacks.js
lundibundi: +1 to @belochub.

lib/callbacks.js
```diff
@@ -3,41 +3,33 @@
...
+const emptiness = () => {};
+
+// Empty asynchronous callback-last single-argument function
+//   callback - <Function>, callback to be called with (null)
+const nop = callback => callback(null);
```
lundibundi: Revert the {} here please.

lib/data.js
```diff
@@ -3,21 +3,21 @@
...
-// Returns: boolean
+//   value - <any>
+// Returns: <boolean>
 const isScalar = value => SCALAR_TYPES.includes(typeof value);
 
 // Copy dataset (copy objects to new array)
-//   ds - array of objects, source dataset
-// Returns: array of objects
+//   ds - <object[]>, source dataset to be copied
+// Returns: <object[]>
```
lundibundi: s/object/Object

lib/data.js
```diff
@@ -192,9 +192,9 @@ const merge = (...args) => {
 };
 
 // Merge multiple objects with merger
-//   merger - function
-//   objs - array of objects
-// Returns: object
+//   merger - <Function>
+//   objs - <object[]>, objects to be merged
```
lundibundi: ditto

lib/events.js
```diff
@@ -21,16 +20,16 @@ const forwardEvent = (
...
-  to, // EventEmitter, to emit event on
-  events // array of strings, events names
-  // Example: common.forwardEvent(from, to);
-  // Example: common.forwardEvent(from, to, 'eventName');
-  // Example: common.forwardEvent(from, to, { eventName: 'newEventName' });
-  // Example: common.forwardEvent(from, to, ['eventName1', 'eventName2']);
-) => {
+// Forward events from one EventEmitter to another
+//   from - EventEmitter, to listen for event
+//   to - EventEmitter, to emit event on
```
lundibundi:
```suggestion
//   to - <EventEmitter>, to emit event on
```
here and above.

lib/events.js
```diff
@@ -58,19 +57,24 @@ util.inherits(EnhancedEmitter, EventEmitter);
...
   forwardEvents(this, to, events);
 };
 
-const emitter = (
-  // Get enhanced EventEmitter with wildcard and forward method
-  // Returns: EventEmitter, instance
-) => (new EnhancedEmitter());
+// Create EnhancedEmitter, enhanced EventEmitter
+// with wildcard and forward method
+// Returns: <Object>, EventEmitter instance
```
lundibundi:
```suggestion
// Returns: <EventEmitter>
```

lib/fp.js
```diff
@@ -2,32 +2,29 @@
...
-const omap = (
-  // Map object fields with provided function
-  mapFn, // funtion, to appy to every field value
-  obj // object, which fields used for mapping
-  // Returns: object, with same reference but with transformed fields
-) => {
+// Partially apply arguments to function
+//   fn - <Function>
+//   args - <Array>, arguments to be applied
+// Returns: <Function>, function(...rest)
```
belochub:
```suggestion
// Returns: <Function>
```

lib/fp.js
```diff
@@ -2,32 +2,29 @@
...
-const compose = (
-  // Compose multiple functions into one
-  ...fns // array of functions
-  // Returns: function, composed
-) => (
-  ...args // arguments to be passed to first function
-) => {
+// Compose multiple functions into one
+//   fns - <Array>, functions to be composed
+// Returns: <Function>, function(...args), composed
```
belochub:
```suggestion
// Returns: <Function>, composed
```

lib/fp.js
```diff
@@ -71,31 +66,29 @@ const zip = (
...
-  // argsI is arguments for i-th currying
-  // argsParts is array of args given for currying from first to i-th currying
-  fn, // function, which will be curried
-  ...args // array, arguments for fn
-  // Returns: function, curried
-) => {
+// Create array of replicated values
+//   count - <number>, new array length
+//   elem - <any>, value to replicate
+// Returns: <array>, replicated
```
belochub:
```suggestion
// Returns: <Array>, replicated
```

lib/fp.js
```diff
@@ -71,31 +66,29 @@ const zip = (
...
+const replicate = (count, elem) => (Array.from({ length: count }, () => elem));
+
+// Zip arrays using specific function
+//   fn - <Function>, for zipping elements with index i
+//   arrays - <Array[]>, arrays to be zipped
+// Returns: <Array>, zipped, element with index i of resulting array is result
+//          of fn called with arguments from arrays
```
belochub: Why is this indented? You haven't indented similar comments in other places.

lib/fp.js
```diff
@@ -71,31 +66,29 @@ const zip = (
...
+const zipWith = (fn, ...arrays) => zip(...arrays).map(args => fn(...args));
+
+// Curry function until the condition is met
+//   condition - <Function>, function(argsI, argsParts) returns boolean
+//     argsI - <Array>, arguments for i-th currying
+//     argsParts - <Array>, of args given for currying
+//                 from first to i-th currying
+//   fn - <Function>, to be curried
+//   args - <Array>, arguments for fn
+// Returns: <Function>, function(...args), curried
```
belochub:
```suggestion
// Returns: <Function>, curried
```

lib/oop.js
```diff
@@ -5,21 +5,19 @@ const inherits = (child, base) => {
...
-const override = (
-  // Override method: save old to `fn.inherited`
-  obj, // object, containing method to override
-  fn // function, name will be used to find method
-  // Hint: Previous function will be accessible by obj.fnName.inherited
-) => {
+// Override method: save old to `fn.inherited`
+//   obj - <Object>, containing method to override
+//   fn - <Function>, name will be used to find method
+// Hint: Previous function will be accessible by obj.fnName.inherited
```
belochub: I think hint should be moved to the description as proposed by @lundibundi.

lib/utilities.js
```diff
@@ -4,11 +4,11 @@ const { basename } = require('path');
...
 const { between } = require('./strings');
 
-const deprecate = (
-  // Wrap method to mark it as deprecated
-  fn // function (optional)
-  // Returns: function, wrapped with deprecation warning
-) => {
+// Wrap method to mark it as deprecated
+//   fn - <Function>
+// Returns: <Function>, function(...args), wrapped with deprecation warning
```
belochub:
```suggestion
// Returns: <Function>, wrapped with deprecation warning
```

lib/utilities.js
```diff
@@ -21,11 +21,11 @@ const deprecate = (
...
 };
 
-const alias = (
-  // Wrap new method to mark old alias as deprecated
-  fn // function (optional)
-  // Returns: function, wrapped with deprecation warning
-) => {
+// Wrap new method to mark old alias as deprecated
+//   fn - <Function>
+// Returns: <Function>,  function(...args), wrapped with deprecation warning
```
belochub:
```suggestion
// Returns: <Function>, wrapped with deprecation warning
```

lib/fp.js
lundibundi: I think we should indent such places everywhere at least to be consistent.

lib/array.js
```diff
@@ -22,7 +22,10 @@ const shuffle = arr => (
 //   to - <number>, range end
 // Returns: <Array>
 //
-// Example: `range(1, 5) = [1, 2, 3, 4, 5]`
+// Example:
+// range(1, 5)
+// Result:
+// [1, 2, 3, 4, 5]
```
lundibundi: Can these at least be on a single line?
```
// Example: range(1, 5)
// Result: [1, 2, 3, 4, 5]
```

SemenchenkoVitaliy: Oh, sorry, I've done this in other places, but skipped this.

lib/fp.js
```diff
@@ -124,37 +116,34 @@ const curry = (
...
+
+// Apply arguments
+//   args - <Array>, arguments to save in closure
+// Returns: <Function>, to be applied saved arguments
+//   args - <Array>, arguments saved in closure
+const applyArgs = (...args) => fn => fn(...args);
+
+// Get first not errored result of fn
+//   fn - <Function>, to be called
+// Returns: <any>, result of `fn(...args)`
```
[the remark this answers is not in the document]
lundibundi: I'd prefer for the nested Result to be indented at the same level as the function.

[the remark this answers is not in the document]
SemenchenkoVitaliy: Than what do you think about
```js
// Get first not errored result of fn
//   fn - <Function>, to be called
// Returns: <Function>
//   args - <Array>, arguments to iterate
//   Returns: <any>, result of `fn(...args)`
// Throws: <Error>, if `fn` throws it
```
so it will be obvious which `Returns` belongs to which function?

README.md
```diff
@@ -11,630 +11,991 @@ Namespace: `api.common` in [Impress Application Server](https://github.com/metar
 ## Installation
 
 ```bash
-$ npm install metasync
+$ npm install metarhia-common
```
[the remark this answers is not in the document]
belochub: I was talking about all of the changes to this file:
>this and other changes to this file

lib/fp.js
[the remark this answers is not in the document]
lundibundi: wdyt of
```
// Get first not errored result of fn
//   fn - <Function>, to be called
// Returns: 
//   <Function>
//     args - <Array>, arguments to iterate
//   Returns: <any>, result of `fn(...args)`
// Throws: <Error>, if `fn` throws it
```
OR
```
// Get first not errored result of fn
//   fn - <Function>, to be called
// Returns: <Function>
//     args - <Array>, arguments to iterate
//   Returns: <any>, result of `fn(...args)`
// Throws: <Error>, if `fn` throws it
```
?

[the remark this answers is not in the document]
belochub: @lundibundi, I, personally, like the second variant more, but in general I am fine with both of them.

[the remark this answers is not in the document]
SemenchenkoVitaliy: As for me, second option looks better.

lib/fp.js
```diff
@@ -124,37 +116,36 @@ const curry = (
...
+// Returns: <Function>
+//   Returns: <Function>, to be applied saved arguments
+//     args - <Array>, arguments saved in closure
+const applyArgs = (...args) => fn => fn(...args);
+
+// Get first not errored result of fn
+//   fn - <Function>, to be called
+// Returns: <Function>
+//     args - <Array>, arguments to iterate
+//   Returns: <any>, result of `fn(...args)`
```
belochub: That's incorrect, it doesn't return the result of `fn(...args)`.

lib/fp.js
```diff
@@ -124,37 +116,36 @@ const curry = (
...
-  ...args // arguments to iterate
-) => {
+// Curry function curry with fn
+//   fn - <Function>, to be curried
+// Returns: <Function>, to pass arguments that returns curried fn
+const curryTwice = fn => curry(curry, fn);
+
+// Apply arguments
+//   args - <Array>, arguments to save in closure
+// Returns: <Function>
```
belochub: You've missed the description of the arguments of this function and added invalid description of the returned value below, the whole description should be:
```js
// Apply arguments
//   args - <Array>, arguments to save in closure
// Returns: <Function>
//     fn - <Function>, to be applied to saved arguments
//   Returns: <any>, result of `fn(...args)`
```

belochub:
Rebase on `master`, please.

lundibundi:
fwiw, why do we need to put commas after the parameter type? Can we make that optional?
(I think this can be fixed with `sed 's/>,/>/g' ./lib`)

belochub:
@SemenchenkoVitaliy have you deliberately omitted [`lib/auth.js`](https://github.com/metarhia/common/blob/comments-refactor/lib/auth.js)?

common.js
```diff
@@ -2,25 +2,25 @@
 
 const submodules = [
   'auth', // Validation of data for authentication/authorization
```
belochub: Shouldn't `auth` be after `array` if that list is sorted alphabetically?

SemenchenkoVitaliy: Yes, sorry.

lib/auth.js
```diff
@@ -140,11 +140,12 @@ const loginPasswordTests = {
 };
 
 class AuthenticationStrength {
-  constructor(
-    valid, // Boolean
-    hints, // Object, { required: Array, optional: Array }
-    compliance // Number, ratio of passed optional tests to all optional tests
-  ) {
+  // AuthenticationStrength constructor
+  //   valid - <boolean>
```
lundibundi: Shouldn't boolean be uppercased too? (refs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean)

lib/auth.js
```diff
@@ -140,11 +140,12 @@ const loginPasswordTests = {
...
+  //   hints - <Object>, { required: Array, optional: Array }
```
lundibundi:
```suggestion
  //   hints - <Object>, { required: <Array>, optional: <Array> }
```

lib/auth.js
SemenchenkoVitaliy: I thought that types should be the same as in nodejs docs. Like [here](https://nodejs.org/api/path.html#path_path_isabsolute_path).

lib/auth.js
SemenchenkoVitaliy: I think this will be better
```js
//   hints - <Object>
//     required - <Array>
//     optional - <Array>
  
```

lib/auth.js
belochub: @lundibundi, no, `boolean` is a scalar type.

lundibundi: Oh, my bad.

lundibundi:
:tada:

PR #195, opened 2018-09-24
Implement count() method for Iterator

belochub:


lib/iterator.js
```diff
@@ -22,6 +22,14 @@ class Iterator {
     return this.base.next();
   }
 
+  size() {
```
nechaido: I'd prefer to name it `amount` or `count`.

nechaido: Maybe even `leftUnconsumed`.

test/iterator.js
```diff
@@ -29,6 +29,21 @@ metatests.test('Iterator is Iterable', test => {
   test.end();
 });
 
+metatests.test('Iterator.size', test => {
+  const iterator = iter(array);
+
+  test.strictSame(iterator.size(), array.length);
```
nechaido: I'd prefer: `iter(array).size()`.

test/iterator.js
```diff
@@ -29,6 +29,21 @@ metatests.test('Iterator is Iterable', test => {
...
+  test.end();
+});
+
+metatests.test('Iterator.size on consumed iterator', test => {
+  const iterator = iter(array);
+  iterator.skip(array.length);
+
+  test.strictSame(iterator.size(), 0);
```
nechaido: I'd prefer: `iter(array).skip(array.length).size()`.

belochub:
Landed in https://github.com/metarhia/common/commit/1c683b2e1723e1c3a64563f64b4a51aa2853da45.

PR #199, opened 2018-10-01
Refactor 'curry' implementation

lundibundi:
* remove dependency on 'curryUntil'/'curryMore'
* use rest syntax

Alternative to https://github.com/metarhia/common/pull/101.

### DIFF
--- a/lib/fp.js
+++ b/lib/fp.js
@@ -113,15 +113,14 @@ const curryUntil = (
 const curry = (
   // Curry function with given arguments
   fn, // function
-  ...args // array, arguments
+  ...par // array, arguments
   // Returns: function, curried
 ) => {
-  let argsTotalCount = 0;
-  const condition = (argsI) => (
-    argsTotalCount += argsI.length,
-    argsTotalCount >= fn.length
-  );
-  return curryUntil(condition, fn, ...args);
+  const curried = (...args) => {
+    par = par.concat(args);
+    return fn.length > par.length ? curried : fn(...par);
+  };
+  return curried();
 };
 
 const curryN = (


aqrln:
I don't think it is okay to mutate `par` this way. Returning a new closure each time the curried function is supplied with new args is the correct behavior, and the only way to optimize it is to introduce a bug.

A simple way to understand why the whole idea (and not even this or other particular implementation) is incorrect: partial application of a function is a new function by definition. There's simply no way for `f: a -> b -> c` and `g: b -> c`  to be the same function, `g = curry(f)` only imposes an additional condition that selects a subset of `Hom(b, c)`, but sets `Hom(a, Hom(b, c))` and `Hom(b, c)` don't intersect at all regardless of that, mathematically speaking. In other words, it is theoretically impossible to achieve what you want to, so the details of implementation aren't important — that would be just different bugs with, maybe, different consequences.

That said, this particular implementation is wrong because it breaks referential transparency and introduces side effects, thus wreaking havoc:

```js
const add = curry(
  (x, y) => x + y
);

const addOne = add(1);
const addTwo = add(2);

console.log(addOne(10));
// Expected `11`, got `3`.

console.log(addTwo(10));
// Expected `12`, got `TypeError: addTwo is not a function`
```

As a side note, speaking about currying itself and not about partial application of curried functions, `f: (a, b) -> c` and `f': a -> b -> c` are equal up to isomorphism, so generally it is possible for `curry(f)` to be the same as `f` — in a language that allows to use curried and uncurried functions interchangeably. `curry` would degenerate into `id` though: `curry = f => f`. JavaScript isn't one of such languages, but it just so happens that a curried function of one argument is a degenerate case of currying anyway, and it is isomorphic through `id` even in JavaScript, so for this particular case you can add an extra check that is guaranteed to be valid:

```js
const curry = (fn, ...par) => {
  if (fn.length === 1 && par.length === 0) {
    return fn;
  }
  // ...
};
```

lundibundi:
Closing in favor of https://github.com/metarhia/common/pull/101. I'll add a few tests there to make sure this will be caught in a testing.

Issue #175, opened 2018-08-27
Flags data type


tshemsedinov:
It is like `Enum` but can store multiple values as a bitmask
```js
const Numeri = new Flags('Uno', 'Due', 'Tre', 'Quattro');
const num = new Flags.from('Due', 'Tre');
const n = num.toNumber();
console.log(n.toString(2)); // "110"
```

tshemsedinov:
Think about compatibility with `Enum`

PR #201, opened 2018-10-03
Add flags data type

SemenchenkoVitaliy:
Closes: https://github.com/metarhia/common/issues/175

### DIFF
--- a/lib/flags.js
+++ b/lib/flags.js
@@ -1,0 +1,89 @@
+'use strict';
+
+const { Uint64 } = require('./uint64');
+
+const masks = [];
+const masksNot = [];
+for (let i = 0; i < 64; i++) {
+  const mask = new Uint64(1).shiftLeft(i);
+  masks.push(mask);
+  masksNot.push(Uint64.not(mask));
+}
+
+const getErrorMsg = key => `Flags instance does not have key ${key}`;
+
+class Flags {
+  static from(...args) {
+    if (args.length > 64) {
+      throw new TypeError('Flags does not support more than 64 values');
+    }
+    const values = new Map(args.map((v, i) => [v, i]));
+
+    class FlagsClass {
+      static from(...args) {
+        return new FlagsClass(...args);
+      }
+
+      constructor(...args) {
+        if (args[0] instanceof Uint64) {
+          this.value = new Uint64(args[0]);
+        } else {
+          this.value = new Uint64(0);
+          args.forEach(arg => this.set(arg));
+        }
+      }
+
+      has(key) {
+        return values.has(key);
+      }
+
+      get(key) {
+        const value = values.get(key);
+        if (value === undefined) {
+          throw new TypeError(getErrorMsg(key));
+        }
+        return (Uint64.shiftRight(this.value, value).toUint32() & 0x1) === 1;
+      }
+
+      set(key) {
+        const value = values.get(key);
+        if (value === undefined) {
+          throw new TypeError(getErrorMsg(key));
+        }
+        this.value.or(masks[value]);
+      }
+
+      unset(key) {
+        const value = values.get(key);
+        if (value === undefined) {
+          throw new TypeError(getErrorMsg(key));
+        }
+        this.value.and(masksNot[value]);
+      }
+
+      toggle(key) {
+        const value = values.get(key);
+        if (value === undefined) {
+          throw new TypeError(getErrorMsg(key));
+        }
+        this.value.xor(masks[value]);
+      }
+
+      toString() {
+        const str = this.value.toString(2);
+        return '0'.repeat(values.size - str.length) + str;
+      }
+
+      toNumber() {
+        return new Uint64(this.value);
+      }
+
+      [Symbol.toPrimitive]() {
+        return this.value.toUint32();
+      }
+    }
+    return FlagsClass;
+  }
+}
+
+module.exports = { Flags };
--- a/test/flags.js
+++ b/test/flags.js
@@ -1,0 +1,242 @@
+'use strict';
+
+const metatests = require('metatests');
+const { Flags, Uint64 } = require('..');
+
+metatests.test('FlagsClass.from', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  test.type(Numeri, 'function');
+
+  const num = Numeri.from('Due', 'Tre');
+
+  test.type(num, 'object');
+  test.strictSame(num.get('Uno'), false);
+  test.strictSame(num.get('Due'), true);
+  test.strictSame(num.get('Tre'), true);
+  test.strictSame(num.get('Quatro'), false);
+
+  test.end();
+});
+
+metatests.test('new FlagsClass', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  test.type(Numeri, 'function');
+
+  const num = new Numeri('Due', 'Tre');
+
+  test.type(num, 'object');
+  test.strictSame(num.get('Uno'), false);
+  test.strictSame(num.get('Due'), true);
+  test.strictSame(num.get('Tre'), true);
+  test.strictSame(num.get('Quatro'), false);
+
+  test.end();
+});
+
+metatests.test('new FlagsClass from Uint64', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  test.type(Numeri, 'function');
+
+  const num = new Numeri(new Uint64('0b10110'));
+
+  test.type(num, 'object');
+  test.strictSame(num.get('Uno'), false);
+  test.strictSame(num.get('Due'), true);
+  test.strictSame(num.get('Tre'), true);
+  test.strictSame(num.get('Quatro'), false);
+
+  test.end();
+});
+
+metatests.test('FlagsClass.get', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  const num = Numeri.from('Due', 'Tre');
+
+  test.strictSame(num.get('Uno'), false);
+  test.strictSame(num.get('Due'), true);
+  test.strictSame(num.get('Tre'), true);
+  test.strictSame(num.get('Quatro'), false);
+
+  test.end();
+});
+
+metatests.test('FlagsClass.has', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  const num = Numeri.from();
+
+  test.strictSame(num.has('Uno'), true);
+  test.strictSame(num.has('Due'), true);
+  test.strictSame(num.has('Tre'), true);
+  test.strictSame(num.has('Quatro'), true);
+
+  test.strictSame(num.has('Cinque'), false);
+  test.strictSame(num.has('Sei'), false);
+
+  test.end();
+});
+
+metatests.test('FlagsClass.set', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  const num = Numeri.from();
+
+  num.set('Due');
+  num.set('Tre');
+
+  test.strictSame(num.get('Due'), true);
+  test.strictSame(num.get('Tre'), true);
+
+  num.set('Due');
+  num.set('Tre');
+
+  test.strictSame(num.get('Due'), true);
+  test.strictSame(num.get('Tre'), true);
+
+  test.end();
+});
+
+metatests.test('FlagsClass.unset', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  const num = Numeri.from('Due', 'Tre');
+
+  num.unset('Due');
+  num.unset('Tre');
+
+  test.strictSame(num.get('Due'), false);
+  test.strictSame(num.get('Tre'), false);
+
+  num.unset('Due');
+  num.unset('Tre');
+
+  test.strictSame(num.get('Due'), false);
+  test.strictSame(num.get('Tre'), false);
+
+  test.end();
+});
+
+metatests.test('FlagsClass.toggle', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  const num = Numeri.from();
+
+  num.toggle('Due');
+  num.toggle('Tre');
+
+  test.strictSame(num.get('Due'), true);
+  test.strictSame(num.get('Tre'), true);
+
+  num.toggle('Due');
+  num.toggle('Tre');
+
+  test.strictSame(num.get('Due'), false);
+  test.strictSame(num.get('Tre'), false);
+
+  test.end();
+});
+
+metatests.test('FlagsClass.toString', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  const nums = [
+    [[], '0000'],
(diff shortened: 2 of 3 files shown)


lib/flags.js
```diff
@@ -0,0 +1,55 @@
...
+class Flags {
+  static from(...args) {
+    const values = args;
+    class FlagsClass {
+      static from(...args) {
+        return new FlagsClass(...args);
+      }
+
+      constructor(...args) {
+        this.values = new Map(values.map(v => [v, 0]));
```
lundibundi: This can be built before this class gets created and captured in scope.

lundibundi: I think the idea was to use Int of another sort of BitPacking in order to store such set values.

lib/flags.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+class Flags {
```
lundibundi: This nesting looks like too much, maybe there is a possibility of extracting the internal class to possibly decrease indentation?

lib/flags.js
```diff
@@ -0,0 +1,55 @@
...
+      constructor(...args) {
+        this.values = new Map(values.map(v => [v, 0]));
+        args.forEach(arg => this.set(arg));
+      }
+
+      has(key) {
+        return this.values.has(key);
+      }
+
+      get(key) {
```
lundibundi: I don't think there is any sense in this method?

lib/flags.js
```diff
@@ -0,0 +1,55 @@
...
+
+      set(key) {
+        if (this.values.has(key)) this.values.set(key, 1);
+      }
+
+      unset(key) {
+        if (this.values.has(key)) this.values.set(key, 0);
+      }
+
+      switch(key) {
```
lundibundi: I'd prefer `toggle` name.

lib/flags.js
```diff
@@ -0,0 +1,55 @@
...
+        if (this.values.has(key))
```
lundibundi: There must not be such checks. All bits are 'unset' by default.

lib/flags.js
```diff
@@ -0,0 +1,55 @@
...
+      toString() {
+        const arr = Array.from(this.values.values());
+        return arr.reduce((acc, val) => val + acc, '');
+      }
+
+      toNumber() {
+        const arr = Array.from(this.values.values());
+        return arr.reduce((acc, val, idx) =>
+          acc | (val << idx), 0);
+      }
```
lundibundi: Missing newline.

lib/flags.js
SemenchenkoVitaliy: Yes, but we need to convert `key` to position of the bit which means that need either to use `Map` or to iterate over array to find this position. There won't be need in `Map`s at all if there was indexes of bits instead of its names.

lib/flags.js
SemenchenkoVitaliy: Inside `Flags.from` `FlagsClass` is created which is binded on context of `Flags.from`, to be more specific - on `args`.

lib/flags.js
SemenchenkoVitaliy: We may need to set bit at a specific position.

lib/flags.js
SemenchenkoVitaliy:
```js
  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
  const num = Numeri.from('Due', 'Tre');
  num.switch('Cinque');
```
The 3rd line shouldn't work, because this key wasn't defined at `Flags.from`.

lib/flags.js
lundibundi: This is not setting a bit, at least it shouldn't. It's a `get` method

lib/flags.js
lundibundi: The third line is incorrect usage, that must not be allowed.
Upon creation (`Flags.from`), you define all of the allowed values.

lib/flags.js
lundibundi: I'm thinking of the way we did it with the `Enum` class. Create a separate function and assign it to `Flags` later (`Flags.from = createFlags` or `Flags.from = (...args) => ...`).

lib/flags.js
lundibundi: Yes, but I think the idea was to have cheap usage (number is already created in the flags instance so `toNumber` is just `return number`) and not so cheap value setting (find value index (we can also use Map for this, no need for linear search) -> set bit at that index).

lib/flags.js
SemenchenkoVitaliy: `Enum` has implementation which is similar to mine and I don't like the idea of assigning one method to empty class. If we have to change it like this I'd rather have just single function. Btw, max level of nesting is 4 which is not so bad.

lib/flags.js
SemenchenkoVitaliy: I agree, but there should be some form of check. I propose to throw error if there is no such defined key. Wdyt?

SemenchenkoVitaliy: Just FYI: `Enum` doesn't need such checks since it doesn't provide interface to change values and is not intended to do this.

lundibundi: Well, it does :smile: https://github.com/metarhia/common/blob/master/lib/enum.js#L14 it cannot change values but you still have to do something when a user asked for an enum instance of a key that doesn't exist.

SemenchenkoVitaliy: Yes, this is a nice proposal :+1: I'll add `NaN` as default value for `get` if key doesn't exist.

lundibundi: I'm not sure, where do you add `NaN`? This is supposed to be BitPacked field I don't think there is any place to put that `NaN`?

lib/flags.js
```diff
@@ -0,0 +1,53 @@
...
+        this.value = 0;
+        args.forEach(arg => this.set(arg));
+      }
+
+      has(key) {
+        return values.has(key);
+      }
+
+      get(key) {
+        return this.value >> values.get(key) & 0x1;
```
belochub: Why are you not validating that `key` provided exists in the `values` Map as you do in other methods?

lib/flags.js
```diff
@@ -0,0 +1,53 @@
...
+      set(key) {
+        if (values.has(key)) this.value |= 0x1 << values.get(key);
+      }
+
+      unset(key) {
+        if (values.has(key)) this.value &= ~(0x1 << values.get(key));
+      }
+
+      toggle(key) {
+        if (values.has(key)) this.get(key) ? this.unset(key) : this.set(key);
```
belochub:
```suggestion
        if (values.has(key)) this.value ^= 1 << values.get(key);
```

lib/flags.js
```diff
@@ -0,0 +1,53 @@
...
+class Flags {
+  static from(...args) {
+    const values = new Map(args.map((v, i) => [v, i]));
+    class FlagsClass {
+      static from(...args) {
+        return new FlagsClass(...args);
+      }
+
+      constructor(...args) {
+        this.value = 0;
```
belochub: I think this implementation should use `Uint64` to store flags instead of the plain `number`; this will increase the total number of flags that can be held in a single type.

lib/flags.js
```diff
@@ -18,19 +20,24 @@ class Flags {
...
       get(key) {
-        return values.has(key) ? this.value >> values.get(key) & 0x1 : -1;
+        return values.has(key) ?
+          Uint64.shiftRight(this.value, values.get(key)).toUint32() & 0x1 :
+          -1;
       }
 
       set(key) {
-        if (values.has(key)) this.value |= 0x1 << values.get(key);
+        if (values.has(key)) this.value.or(
```
lundibundi: We don't use multiline if without braces, could you please add {}?

lib/flags.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+      }
+
+      toString() {
+        const str = this.value.toString(2);
+        return '0'.repeat(args.length - str.length) + str;
+      }
+
+      toNumber() {
+        return this.value.toUint32();
```
belochub: I think this should return `Uint64` as well.

lib/flags.js
```diff
@@ -0,0 +1,63 @@
...
+
+class Flags {
+  static from(...args) {
+    const values = new Map(args.map((v, i) => [v, i]));
+    class FlagsClass {
+      static from(...args) {
+        return new FlagsClass(...args);
+      }
+
+      constructor(...args) {
```
belochub: Also, I think it is important to implement support for construction of the `FlagsClass` object from `Uint64` number (since that's how it is going to be stored).
At the moment that is only achievable via changing the `value` field of the resulting object, which isn't very convenient.

lib/flags.js
```diff
@@ -0,0 +1,67 @@
...
+
+      toggle(key) {
+        if (values.has(key)) {
+          this.value.xor(new Uint64(1).shiftLeft(values.get(key)));
+        }
+      }
+
+      toString() {
+        const str = this.value.toString(2);
+        return '0'.repeat(args.length - str.length) + str;
```
belochub:
```suggestion
        return '0'.repeat(values.size - str.length) + str;
```

lib/flags.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+
+      has(key) {
+        return values.has(key);
+      }
+
+      get(key) {
+        return values.has(key) ?
+          Uint64.shiftRight(this.value, values.get(key)).toUint32() & 0x1 :
+          -1;
```
tshemsedinov: to previous line

lib/flags.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+
+      set(key) {
+        if (values.has(key)) {
+          this.value.or(new Uint64(1).shiftLeft(values.get(key)));
+        }
+      }
+
+      unset(key) {
+        if (values.has(key)) {
```
lundibundi: So if the user provided invalid value we will just silently ignore and proceed?
e.g. we have flags of ('Uno', 'Due', 'Tre') and user writes `flags.toggle('Duo')` which will just ignore the call without any error.

I think all such cases should result in TypeError.
/cc @belochub @nechaido @tshemsedinov

lib/flags.js
```diff
@@ -0,0 +1,66 @@
...
+          this.value.and(new Uint64(1).shiftLeft(values.get(key)).not());
```
lundibundi: Cache all Uints (0b1, 0b10, 0b100 ...) and get needed from array? =)

test/flags.js
```diff
@@ -0,0 +1,194 @@
+'use strict';
+
+const metatests = require('metatests');
+const { Flags, Uint64 } = require('..');
+
+metatests.test('FlagsClass.from', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  test.strictSame(typeof Numeri, 'function');
```
lundibundi: There is `test.type`.

lib/flags.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+      }
+
+      has(key) {
+        return values.has(key);
+      }
+
+      get(key) {
+        return values.has(key) ?
+          Uint64.shiftRight(this.value, values.get(key)).toUint32() & 1 : -1;
```
lundibundi: Wouldn't it be better to return true/false (set/not-set) from this method?

lib/flags.js
SemenchenkoVitaliy: That will mean that we will duplicate Uint in the array. Also, if we are to duplicate it I'd prefer to use another `Map`. But as far as I know, `Flags` should have fast read speed meaning that data should be written on value change.

lib/flags.js
SemenchenkoVitaliy: I agree but we should handle incorrect `key`s, though if we will throw errors in `set`, `unset` and `toggle` than I'm fine with throwing error here too. Also, it primarily depends on how `Flags` will be used and whether there will be used `Number` or `boolean`.

lib/flags.js
lundibundi: What do you mean
> duplicate Uint in the array

?
I suggested to have an array of format `[UIn64(0b0), UIn64(0b1), UIn64(0b10), UIn64(0b100)...]`
And then use `this.value.and(arr[value.get(key)]).not()` etc.

lib/flags.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+const { Uint64 } = require('./uint64');
+
+class Flags {
+  static from(...args) {
+    const values = new Map(args.map((v, i) => [v, i]));
```
lundibundi: Could you enforce that args.length is <= 64?

lib/flags.js
```diff
@@ -0,0 +1,80 @@
...
+      get(key) {
+        return values.has(key) ?
+          Uint64.shiftRight(this.value, values.get(key)).toUint32() & 1 : -1;
+      }
+
+      set(key) {
+        if (values.has(key)) {
+          this.value.or(masks[values.get(key)]);
+        } else {
+          throw new TypeError('Flags instance does not have such key');
```
belochub: You can invert the condition and leave only one `if` branch here and in a bunch of other methods:
```js
if (!values.has(key)) {
  throw new TypeError(/*...*/);
}
this.value.operation(/*...*/);
```

lundibundi: Could you extract `get` and have it:
```js
const v = values.get(key);
if (v === undefined) {
  throw ...
}
this.value.and(masks[v]).not();
```

lib/flags.js
lundibundi: Ditto https://github.com/metarhia/common/pull/201/files#r232201310.

lib/flags.js
lundibundi: Ok, let's make it return true/false and throw if `key` is not present.

/cc @belochub @nechaido

SemenchenkoVitaliy:
@lundibundi There are tests for all cases that may inflict error throwing. Or do we need a few for each one?

lib/flags.js
```diff
@@ -0,0 +1,87 @@
...
+        if (v === undefined) {
+          throw new TypeError('Flags instance does not have such key');
+        }
+        this.value.and(masksNot[v]);
+      }
+
+      toggle(key) {
+        const v = values.get(key);
+        if (v === undefined) {
+          throw new TypeError('Flags instance does not have such key');
```
lundibundi: Add `key` to the message please.

test/flags.js
```diff
@@ -0,0 +1,242 @@
...
+  test.end();
+});
+
+metatests.test('Toggle incorrect key', test => {
+  const Numeri = Flags.from('Uno', 'Due', 'Tre', 'Quatro');
+  const num = Numeri.from();
+  const message = 'Flags instance does not have key Cinque';
+  test.throws(() => {
+    num.toggle('Cinque');
+  }, new Error(message));
```
belochub: We should probably test for `TypeError` here:
```suggestion
  }, new TypeError(message));
```

lib/flags.js
```diff
@@ -0,0 +1,87 @@
...
+      }
+
+      has(key) {
+        return values.has(key);
+      }
+
+      get(key) {
+        const v = values.get(key);
+        if (v === undefined) {
+          throw new TypeError(`Flags instance does not have key ${key}`);
```
belochub: Should we maybe extract the template literal for error message into the function since it is used in multiple places?

belochub:
Landed in https://github.com/metarhia/common/commit/6b7c13da685bc7a4b875eb3f33b06da9c49e4be9.

PR #203, opened 2018-10-12
Implement Int64

SemenchenkoVitaliy:


lib/int64.js
```diff
@@ -0,0 +1,329 @@
...
+      this.value[0] = res.value[0];
+      this.value[1] = res.value[1];
+    }
+  }
+
+  toInt32() {
+    return this.value[0];
+  }
+
+  static _convertion(value) {
```
belochub: Should probably be `_conversion`.

test/int64.js
```diff
@@ -0,0 +1,529 @@
...
+  test.throws(() => {
+    number.toString(0);
+  }, errorExpected, 'must throw when toString() radix argument is too small');
+  test.throws(() => {
+    number.toString(-1);
+  }, errorExpected, 'must throw when toString() radix argument is negative');
+
+  test.end();
+});
+*/
```
tshemsedinov: Why do we have 162 commented lines here?

SemenchenkoVitaliy: That's the tests that will be fixed in the next commit.

lib/int64.js
```diff
@@ -0,0 +1,335 @@
...
+        this.value[1] = 0xFFFFFFFF;
+      } else {
+        this.value[1] = 0;
+      }
+    } else {
+      const mask = (1 << b) - 1;
+      const tr = (this.value[1] & mask) << (32 - b);
+      this.value[0] = (this.value[0] >>> b) | tr;
+      this.value[1] = this.value[1] >>> b;
+      if (sign) this.value[1] |= mask << (32 - b);
```
belochub: I think you can remove this line if you use `>>` operator instead of `>>>` on the previous line.

lib/int64.js
```diff
@@ -0,0 +1,335 @@
...
+    if (b >= 32) {
+      this.value[1] = this.value[0] << (b - 32);
+      this.value[0] = 0;
+    } else {
+      const mask = ((1 << b) - 1) << (32 - b);
+      const tr = (this.value[0] & mask) >>> (32 - b);
+      this.value[0] = this.value[0] << b;
+      this.value[1] = (this.value[1] << b) | tr;
+    }
+    if (sign) this.value[1] |= 0x80000000;
```
belochub: This behavior is incorrect, the sign must not be frozen (if I understand correctly that it's what happens here), when shifting left sign bit may become 0.

Also, you should add some more tests cases on negative numbers shifting since there are only 2 of them now, and they both use -1 as a left operand.

test/int64.js
```diff
@@ -0,0 +1,663 @@
...
+      expectedString: '9223372036854775807',
+      message: 'must create Int64 from a big oct number provided as a string',
+    },
+    {
+      value: '9223372036854775807',
+      expectedString: '9223372036854775807',
+      message: 'must create Int64 from a big dec number provided as a string',
+    },
+    {
+      value: '-9223372036854775807',
```
belochub: Why not use `-9223372036854775808` here as the minimal int64 value, similar to the test case above, where you use the maximal int64 value?

lib/int64.js
```diff
@@ -0,0 +1,335 @@
...
+
+  static _conversion(value) {
+    const result = new Uint32Array(value);
+    let pos = 0;
+    if (result[0] === 0) {
+      while (pos < 32) {
+        if ((result[1] >>> pos) & 1) break;
+        pos++;
+      }
+      const mask = (1 << (pos + 1)) - 1;
```
lundibundi: This will potentially shift by 32 (max pos in 31 and then +1) which will result in `1` and therefore `(1 << (pos + 1)) - 1` will be `0`, is that intentional?

belochub: This also seems to lead to a bug when handling value mentioned in the https://github.com/metarhia/common/pull/203#discussion_r227176378.

lib/int64.js
```diff
@@ -0,0 +1,335 @@
...
+      this.value[0] = res.value[0];
+      this.value[1] = res.value[1];
+    }
+  }
+
+  toInt32() {
+    return this.value[0];
+  }
+
+  static _conversion(value) {
```
lundibundi: IIUC this is twos complement conversion (negative number to positive number representation), could you name as such to be clear?

lib/int64.js
```diff
@@ -0,0 +1,335 @@
...
+
+  static _division(n, d) {
+    const zero = new Int64();
+    const one = new Int64(1);
+    if (Int64.cmp(d, zero) === 0) {
+      throw new RangeError('Int64: division by zero');
+    }
+    if (Int64.cmp(n, zero) === 0) return [zero, 0];
+    const cmp = Int64.cmp(d, n);
+    if (cmp === 0) return [one, zero];
```
lundibundi: I think this should also check for different-sign division i.e. -10/10, 10/-10.

lib/int64.js
```diff
@@ -0,0 +1,335 @@
...
+
+    const nSign = n.value[1] >>> 31;
+    let value = new Uint32Array(n.value);
+    if (nSign) value = Int64._conversion(value);
+    const dSign = d.value[1] >>> 31;
+    const dPositive = new Int64(d);
+    if (dSign) dPositive.value = Int64._conversion(dPositive.value);
```
lundibundi: IIUC this is converting negative num to positive number representation in order to divide them as unsigned. If that’s the case I’d want to actually implement signed-division instead of converting them here and at the end.

lib/int64.js
```diff
@@ -0,0 +1,335 @@
...
+    if (Int64.cmp(a, zero) === 0 || Int64.cmp(b, zero) === 0) return zero;
+    if (Int64.cmp(a, one) === 0) return new Int64(b);
+    if (Int64.cmp(b, one) === 0) return new Int64(a);
+
+    const aSign = a.value[1] >>> 63;
+    const bSign = b.value[1] >>> 63;
+    const result = new Int64();
+    const aPositive = new Int64(a);
+    let value = new Uint32Array(b.value);
+    if (aSign) aPositive.value = Int64._conversion(aPositive.value);
```
lundibundi: https://github.com/metarhia/common/pull/203/files?utf8=%E2%9C%93&diff=unified#r227758289 Also applies here, I think there is better way of multiplying these than transforming them twice.

lib/int64.js
```diff
@@ -0,0 +1,355 @@
...
+        value = value.slice(2);
+        const val0 = Number('0o' + value.slice(value.length - 11));
+        const val1 = Number('0o' + value.slice(0, value.length - 11));
+        this.value[0] = val0;
+        this.value[1] = (val1 << 1) | (val0 >>> 31);
+      } else if (value.startsWith('0b')) {
+        value = value.slice(2);
+        this.value[0] = Number('0b' + value.slice(value.length - 32));
+        this.value[1] = Number('0b' + value.slice(0, value.length - 32));
+      } else {
```
lundibundi: Those conversions won't work if the string  is longer than supported by Number. I'd suggest to just use the way it was done in Uint64.

lib/int64.js
```diff
@@ -0,0 +1,373 @@
...
+      this.value[1] = UINT32_MAX;
+    } else if (numValue >= 0 && numValue <= Number.MAX_SAFE_INTEGER) {
+      this.value[0] = numValue % (UINT32_MAX + 1);
+      this.value[1] = Math.floor(numValue / (UINT32_MAX + 1));
+    } else if (numValue >= Number.MIN_SAFE_INTEGER && numValue < 0) {
+      this.value[0] = -numValue % (UINT32_MAX + 1);
+      this.value[1] = Math.floor(-numValue / (UINT32_MAX + 1));
+      this.value = Int64._conversion(this.value);
+    } else if (typeof value === 'string') {
+      let sign = false;
```
lundibundi: I'd prefer to have more explicit name, i.e. `negative`/`neg`?

lib/int64.js
```diff
@@ -0,0 +1,373 @@
...
+        if (sign) res = Int64.mult(res, intRadix).sub(intDigit);
+        else res = Int64.mult(res, intRadix).add(intDigit);
+      }
+      this.value[0] = res.value[0];
+      this.value[1] = res.value[1];
+    }
+  }
+
+  toInt32() {
+    return this.value[0];
```
lundibundi: Is this correct for negative numbers?

lib/int64.js
```diff
@@ -0,0 +1,373 @@
...
+    return this;
+  }
+
+  static add(a, b) {
+    return new Int64(a).add(b);
+  }
+
+  sub(b) {
+    let value = new Uint32Array(b.value);
+    if (value[0] !== 0 || value[1] !== 0) value = Int64._conversion(value);
```
lundibundi: IIUC maybe replace this with
```js
if (b.value[0] === 0 && b.value[1] === 0) return this;
const value = Int64._conversion(b.value);
```
?

lib/int64.js
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  static sub(a, b) {
+    return new Int64(a).sub(b);
+  }
+
+  static cmp(a, b) {
+    const aSign = a.value[1] >>> 31;
+    const bSign = b.value[1] >>> 31;
+    if (aSign !== bSign) return aSign ? -1 : 1;
```
lundibundi:
```suggestion
    if (aSign !== bSign) return aSign === 1 ? -1 : 1;
```
?

lib/int64.js
```diff
@@ -0,0 +1,373 @@
...
+
+  static mult(a, b) {
+    const zero = new Int64();
+    const one = new Int64(1);
+    if (Int64.cmp(a, zero) === 0 || Int64.cmp(b, zero) === 0) return zero;
+    if (Int64.cmp(a, one) === 0) return new Int64(b);
+    if (Int64.cmp(b, one) === 0) return new Int64(a);
+
+    const aSign = a.value[1] >>> 63;
+    const bSign = b.value[1] >>> 63;
```
lundibundi:
```suggestion
    const bSign = b.value[1] >>> 31;
```
?

lib/int64.js
tshemsedinov: `negative === true` if value `< 0`
I'd propose `positive === true` if value `> 0`

lib/int64.js
```diff
@@ -0,0 +1,377 @@
...
+    const mask = (1 << b) - 1;
+    const tr = (this.value[1] & mask) << (32 - b);
+    this.value[1] = this.value[1] >>> b;
+    this.value[0] = (this.value[0] >>> b) | tr;
+    return this;
+  }
+
+  shiftRightArithmetic(b) {
+    b %= 63;
+    const sign = this.value[1] >>> 31;
```
lundibundi: Also, neg/pos?

lib/int64.js
```diff
@@ -0,0 +1,377 @@
...
+  inc() {
+    return this.add(new Int64(1));
+  }
+
+  dec() {
+    return this.sub(new Int64(1));
+  }
+
+  toString(radix = 10) {
+    const sign = this.value[1] >>> 31;
```
lundibundi: Ditto.

lib/int64.js
```diff
@@ -0,0 +1,377 @@
+'use strict';
+
+const { Uint64 } = require('./uint64');
+
+const INT32_MAX = 0x7FFFFFFF;
+const UINT32_MAX = 0xFFFFFFFF;
+
+const charToNum = {
+  0: 0, 1: 1, 2: 2,
```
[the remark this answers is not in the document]
SemenchenkoVitaliy: If we are going to change it, I'd say that we should also change the same thing in [Uint64](https://github.com/metarhia/common/blob/master/lib/uint64.js#L5).

lib/int64.js
```diff
@@ -0,0 +1,377 @@
...
+    if (aSign !== bSign) return aSign === 1 ? -1 : 1;
+
+    if (a.value[1] > b.value[1]) return 1;
+    if (a.value[1] < b.value[1]) return -1;
+    if (a.value[0] === b.value[0]) return 0;
+    return a.value[0] > b.value[0] ? 1 : -1;
+  }
+
+  static _division(n, d) {
+    const zero = new Int64();
```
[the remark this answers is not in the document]
SemenchenkoVitaliy: I'd prefer to have something like
```js
Int64.getZero() {
  return new Int64(0);
}
```
otherwise, every operation that modifies value should be done like this
```js
const zero = new Int64(Int64.zero);
```
so the value won't be overwritten.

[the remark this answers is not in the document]
lundibundi: +1, but maybe just `Int64.zero()`?

lundibundi:
Yeah, that's a given =)

lib/int64.js
```diff
@@ -0,0 +1,385 @@
...
+        res = Int64.mult(res, intRadix);
+        if (negative) res.sub(intDigit);
+        else res.add(intDigit);
+      }
+      this.value[0] = res.value[0];
+      this.value[1] = res.value[1];
+    }
+  }
+
+  toInt32() {
```
belochub: This method should just return `this.value[0]`.

lib/int64.js
```diff
@@ -0,0 +1,385 @@
...
+    }
+    const mask = (1 << b) - 1;
+    const tr = (this.value[1] & mask) << (32 - b);
+    this.value[1] = this.value[1] >>> b;
+    this.value[0] = (this.value[0] >>> b) | tr;
+    return this;
+  }
+
+  shiftRightArithmetic(b) {
+    b %= 63;
```
belochub: Why do you use 63 and not 64 here?
```suggestion
    b %= 64;
```

lib/int64.js
SemenchenkoVitaliy: I thought that if the value is negative we should also return a negative number. For example, now it converts like that:
```js
new Int64('-1').toInt32() -> -1
```
If we just return `value[0]` it will be like this:
```js
new Int64('-1').toInt32() -> 4294967295
```
Do we need such type of behaviour?

lib/int64.js
```diff
@@ -0,0 +1,385 @@
...
+  c: 12, C: 12,
+  d: 13, D: 13,
+  e: 14, E: 14,
+  f: 15, F: 15,
+};
+
+class Int64 {
+  constructor(value) {
+    this.value = new Uint32Array(2);
+    if (value instanceof Int64) {
```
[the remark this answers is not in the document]
SemenchenkoVitaliy: But what if the value will be higher than `0x7FFFFFFFFFFFFFFF`? Should I add some kind of a check for such case?

lib/int64.js
belochub: Yeah, I'm sorry, you should return `this.value[0] >> 0` here.
Separate method called `toUint32()` that returns just `this.value[0]` can be useful here, though.

lib/int64.js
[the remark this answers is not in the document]
belochub: No, there is no need to check anything, this will be a conversion preserving the exact bit-value of the numbers.

belochub:
Landed in https://github.com/metarhia/common/commit/421594343984d0738f4d138c07f1c105a9547b25.

PR #209, opened 2018-10-31
Implement Iterator.enumerate to return [index, value]

lundibundi:


belochub:
Landed in https://github.com/metarhia/common/commit/d62655536e1580c82d3f86b0de13cae402e72fff.

Issue #213, opened 2018-11-02
Array can't be processed by duplicateObject


tshemsedinov:
```
/home/travis/build/metarhia/metaschema/node_modules/metarhia-common/lib/data.js:108
      object[key] = val;
                  ^
TypeError: Cannot assign to read only property 'length' of object '[object String]'
    at duplicateObject (/metaschema/node_modules/metarhia-common/lib/data.js:108:19)
```
See CI output: https://travis-ci.org/metarhia/metaschema/jobs/449710673

PR #214, opened 2018-11-02
Fix `duplicate()` for Objects created in different contexts

belochub:
Fixes: https://github.com/metarhia/common/issues/213


tshemsedinov:
Landed in df55911

PR #216, opened 2018-11-05
Update Travis config to use the current Node

belochub:
Remove end-of-life Node.js 9 and add the current Node.js 11.


belochub:
Landed in https://github.com/metarhia/common/commit/53f7e075057696291fc3e594b73f0c5847f48df6.

Issue #205, opened 2018-10-19
Identity functions


tshemsedinov:
Functional programming classic Identity is:
```js
const id = x => x;
```
but we need async Identity, returning initial value to callback second argument:
```js
const asyncId => (x, callback) => {
  callback(null, x);
};
```
Array iteration functions contract is `(array, fn, callback)` so we need something like this:
```js
const arrayIterationId = (items, fn, callback) => {
  callback(null, items.map(id));
};
```
For example we need it here: https://github.com/metarhia/metasync/pull/379/files#diff-8359eeb9ceb67853754febf281ddb569R14

We may place all Identity implementations in `common` or place sync implementation in `common` and async implementation in `metasync`.

tshemsedinov:
@nechaido @belochub please assign comebody

machendos:
Should I change all the old uses of similar functions to a new implementation in ``common`` in this issue?

PR #217, opened 2018-11-10
Implement identity functions

machendos:
Closed:https://github.com/metarhia/common/issues/205

### DIFF
--- a/lib/callbacks.js
+++ b/lib/callbacks.js
@@ -96,7 +96,7 @@ const safeFunction = fn => (typeof fn === 'function' ? fn : emptiness);
 const id = x => x;
 
 // Async identity function
-//   x - <any>, incoming value which will be returned in callback
+//   x - <any>, incoming value which will be returned into the callback
 //   callback - <Function>, callback to be called with first argument
 //     err - <null>
 //     data - <any>
--- a/lib/callbacks.js
+++ b/lib/callbacks.js
@@ -90,6 +90,19 @@ const unsafeFunction = fn => (typeof fn === 'function' ? fn : null);
 // Returns: <Function>, function or `common.emptiness` if fn is not a function
 const safeFunction = fn => (typeof fn === 'function' ? fn : emptiness);
 
+// Identity function
+//   x - <any>, incoming value which will be returned
+// Returns: <any>, incoming value
+const id = x => x;
+
+// Async identity function
+//   x - <any>, incoming value which will be returned in callback
+//   callback - <Function>, callback to be called with first argument
+const asyncId = (x, callback) => {
+  process.nextTick(callback, null, x);
+};
+
+
 module.exports = {
   falseness,
   trueness,
@@ -112,4 +125,7 @@ module.exports = {
 
   safeFunction,
   unsafeFunction,
+
+  id,
+  asyncId,
 };
--- a/lib/callbacks.js
+++ b/lib/callbacks.js
@@ -98,8 +98,8 @@ const id = x => x;
 // Async identity function
 //   x - <any>, incoming value which will be returned in callback
 //   callback - <Function>, callback to be called with first argument
-//     err - <null>, will be always return instead of error
-//     data - <any>, incoming value
+//     err - <null>
+//     data - <any>
 const asyncId = (x, callback) => {
   process.nextTick(callback, null, x);
 };
--- a/lib/callbacks.js
+++ b/lib/callbacks.js
@@ -98,7 +98,7 @@ const id = x => x;
 // Async identity function
 //   x - <any>, incoming value which will be returned in callback
 //   callback - <Function>, callback to be called with first argument
-//     err - <null>, will be alse return instead of error
+//     err - <null>, will be always return instead of error
 //     data - <any>, incoming value
 const asyncId = (x, callback) => {
   process.nextTick(callback, null, x);
--- a/lib/callbacks.js
+++ b/lib/callbacks.js
@@ -98,6 +98,8 @@ const id = x => x;
 // Async identity function
 //   x - <any>, incoming value which will be returned in callback
 //   callback - <Function>, callback to be called with first argument
+//     err - <null>, will be alse return instead of error
+//     data - <any>, incoming value
 const asyncId = (x, callback) => {
   process.nextTick(callback, null, x);
 };
--- a/lib/fp.js
+++ b/lib/fp.js
@@ -186,14 +186,6 @@ const asyncId = (x, callback) => {
   callback(null, x);
 };
 
-// Mapable object iteration identity function
-//   items - <Array> | <Object> with map method
-//   fn - <any>, Not used. Reserved for compatibility with other iterator func
-//   callback - <Function>, will be called with copy of items
-const mapIterationId = (items, fn, callback) => {
-  callback(null, items.map(id));
-};
-
 module.exports = {
   partial,
   omap,
--- a/lib/fp.js
+++ b/lib/fp.js
@@ -175,21 +175,21 @@ const restLeft = fn => (...spreadArgs) => {
 };
 
 // Identity function
-//   x - any argument
-// Result: argument x
+//   x - <any>, incoming value which will be returned
+// Returns: <any>, incoming value
 const id = x => x;
 
 // Async identity function
-//   x - any argument
-//   callback - function, (err, data)
+//   x - <any>, incoming value which will be returned in callback
+//   callback - <Function>, callback to be called with first argument
 const asyncId = (x, callback) => {
   callback(null, x);
 };
 
 // Mapable object iteration identity function
-//   items - array
-//   fn - function
-//   callback - function, (err, data)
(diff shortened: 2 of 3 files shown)


lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
...
+const ident = x => x;
+
+// Async identity function
+//   x - any argument
+//   callback - function, (err, data)
+const asyncIdent = (x, callback) => {
+  callback(null, x);
+};
+
+// Array Iteration identity function
```
tshemsedinov: Not only Array, any mapable object

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
...
+//   items - array
+//   fn - function
+//   callback - function, (err, data)
+const arrayIterationIdent = (items, fn, callback) => {
```
tshemsedinov:
```suggestion
const mapIterationId = (items, fn, callback) => {
```

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
...
 
+// Identity function
+//   x - any argument
+// Result: argument x
+const ident = x => x;
+
+// Async identity function
+//   x - any argument
+//   callback - function, (err, data)
+const asyncIdent = (x, callback) => {
```
tshemsedinov:
```suggestion
const asyncId = (x, callback) => {
```

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
   fn(args, ...namedArgs, callback);
 };
 
+// Identity function
+//   x - any argument
+// Result: argument x
+const ident = x => x;
```
tshemsedinov:
```suggestion
const id = x => x;
```

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
...
+//   callback - function, (err, data)
+const asyncId = (x, callback) => {
+  callback(null, x);
+};
+
+// Mapable object iteration identity function
+//   items - array
+//   fn - function
+//   callback - function, (err, data)
+const mapIterationId = (items, fn, callback) => {
```
nechaido: I see no reason behind this function.

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
...
+  callback(null, items.map(items));
```
nechaido:
```suggestion
  callback(null, items.map(fn));
```

machendos: It's identify iteration function. So, I think, for every element we have call identify function

lib/fp.js
belochub: You don't seem to be using the second `fn` argument, why have you added it to the function's signature?

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
...
+const asyncId = (x, callback) => {
+  callback(null, x);
+};
+
+// Mapable object iteration identity function
+//   items - array
+//   fn - function
+//   callback - function, (err, data)
+const mapIterationId = (items, fn, callback) => {
+  callback(null, items.map(id));
```
belochub: In this case, `Array.prototype.map()` can be replaced with `Array.prototype.slice()`, this won't modify the behavior:
```suggestion
  callback(null, items.slice());
```

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
   fn(args, ...namedArgs, callback);
 };
 
+// Identity function
+//   x - any argument
```
belochub: You should use the new documentation comments style, there are plenty of examples of the correct style in this project since all of the comments were updated in https://github.com/metarhia/common/pull/194.

lib/fp.js
machendos: Considering this, I think there can be more than just arrays. Perhaps any objects with the map method: https://github.com/metarhia/common/pull/217#discussion_r232493540

lib/fp.js
machendos: I thought there should be the same signature as in other functions, iterating over arrays

lib/fp.js
tshemsedinov: How can we describe "any type" or scalar? I thin here we have scalar type

lib/fp.js
nechaido: ping. What is the reason behind this function?

machendos: Perhaps it would be convenient to return this function where the iterator function is expected, however we have nothing to give

lib/fp.js
```diff
@@ -174,6 +174,26 @@ const restLeft = fn => (...spreadArgs) => {
...
+// Identity function
+//   x - <any>, incoming value which will be returned
+// Returns: <any>, incoming value
+const id = x => x;
+
+// Async identity function
+//   x - <any>, incoming value which will be returned in callback
+//   callback - <Function>, callback to be called with first argument
+const asyncId = (x, callback) => {
+  callback(null, x);
```
belochub: It isn't really async now, you should call the callback in `process.nextTick()`:
```suggestion
  process.nextTick(callback, null, x);
```

lib/fp.js
tshemsedinov: We discussed this and agreed to remove function

SemenchenkoVitaliy:
As for me, I'd prefer it to be in `common/callbacks.js` instead of `common/fp.js`since it does not process arguments though I don't mind either way.

lib/callbacks.js
```diff
@@ -90,6 +90,19 @@ const unsafeFunction = fn => (typeof fn === 'function' ? fn : null);
...
 const safeFunction = fn => (typeof fn === 'function' ? fn : emptiness);
 
+// Identity function
+//   x - <any>, incoming value which will be returned
+// Returns: <any>, incoming value
+const id = x => x;
+
+// Async identity function
+//   x - <any>, incoming value which will be returned in callback
+//   callback - <Function>, callback to be called with first argument
```
nechaido: You should document callback arguments.

lib/callbacks.js
```diff
@@ -90,6 +90,21 @@ const unsafeFunction = fn => (typeof fn === 'function' ? fn : null);
...
+//     err - <null>, will be alse return instead of error
```
nechaido: I can't quite get the meaning of this.

nechaido: I think the description of callback arguments can be omitted, yet types should stay.

lib/callbacks.js
```diff
@@ -90,6 +90,21 @@ const unsafeFunction = fn => (typeof fn === 'function' ? fn : null);
 // Returns: <Function>, function or `common.emptiness` if fn is not a function
 const safeFunction = fn => (typeof fn === 'function' ? fn : emptiness);
 
+// Identity function
+//   x - <any>, incoming value which will be returned
+// Returns: <any>, incoming value
+const id = x => x;
+
+// Async identity function
+//   x - <any>, incoming value which will be returned in callback
```
nechaido:
```suggestion
//   x - <any>, incoming value which will be returned into the callback
```

lib/callbacks.js
```diff
@@ -90,6 +90,21 @@ const unsafeFunction = fn => (typeof fn === 'function' ? fn : null);
...
+//   x - <any>, incoming value which will be returned
+// Returns: <any>, incoming value
+const id = x => x;
+
+// Async identity function
+//   x - <any>, incoming value which will be returned into the callback
+//   callback - <Function>, callback to be called with first argument
+//     err - <null>
+//     data - <any>
+const asyncId = (x, callback) => {
```
lundibundi: I'm not sure about the usage of this function. I thought this would be some sort of callback-identity to be used when there is no need in a specific callback but the implementation requests one or there are multiple arguments returned but you only need the first one.
i.e.
```js
const asyncId = callback => (err, x) => {
  if (err) callback(err);
  else callback(null, x);
};
```

test/identify.js
```diff
@@ -0,0 +1,20 @@
...
+});
+
+metatests.test('async identify function', test => {
+  const cb = (err, data) => {
+    test.error(err);
+    test.strictSame(data, { data: 10 });
+    test.end();
+  };
+
+  common.asyncId({ data: 10 }, cb);
```
lundibundi: This doesn't check that `asyncId` is indeed async. You should
```js
  let sync = true;
  const cb = (err, data) => {
    test.error(err);
    test.assertNot(sync);
    test.strictSame(data, { data: 10 });
    test.end();
  };
  common.asyncId({ data: 10 }, cb);
  sync = false;
```
or something like it.

belochub:
Landed in https://github.com/metarhia/common/commit/7eab27574177a275b53ce36f5e04eb460b529f22.

PR #221, opened 2018-11-14
Fix incorrect behavior in Uint64 constructor

belochub:
It resulted in invalid values being created when passing a binary
number with digit count between 53 and 63 to the constructor as a
string.
Also, refactor and optimize creation of values from strings representing
numbers in other numeral systems.

### DIFF
--- a/lib/uint64.js
+++ b/lib/uint64.js
@@ -15,6 +15,12 @@ const charToNum = {
   f: 15, F: 15,
 };
 
+const radixStr = {
+  '0x': 16,
+  '0o': 8,
+  '0b': 2,
+};
+
 class Uint64 {
   constructor(value) {
     this.value = new Uint32Array(2);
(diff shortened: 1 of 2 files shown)


lib/uint64.js
```diff
@@ -43,7 +43,7 @@ class Uint64 {
       }
       const uintRadix = new Uint64(radix);
       let res = new Uint64();
-      for (let i = 0; i < value.length; i++) {
+      for (let i = radix !== 2 ? 0 : 2; i < value.length; i++) {
```
lundibundi:
```suggestion
      for (let i = radix === 10 ? 0 : 2; i < value.length; i++) {
```
?

belochub: I've done it the way you propose at first, but then I've changed it to the way it is now since it is shorter and results in the same behavior.

belochub:
Landed in https://github.com/metarhia/common/commit/8257345d227d4f18a7a81d7bcd72ca39bf0a61c7.

Issue #138, opened 2018-07-12
Rename module


tshemsedinov:
We need shorter and conceptual name for npm. Now npm module name is `metarhia-common`

nechaido:
@tshemsedinov I personally prefer the current name.

tshemsedinov:
@nechaido I'd like to have same names in npm and git

tshemsedinov:
`metalib` ?

nechaido:
@tshemsedinov sounds ok, but `@metarhia/common` is the best option IMHO.

tshemsedinov:
@nechaido but thus we should move all libs to https://www.npmjs.com/org/metarhia

nechaido:
@tshemsedinov I would like that.

PR #222, opened 2018-11-16
Version 1.0.0

belochub:
* Rename package to @metarhia/common and update all the mentions in
  README.md.
* Update devDependencies.
* Use "files" field in package.json instead of .npmignore.
* Publish lib/unicode-categories.js to npm instead of generating it on
  install since it was published inside dist/ anyway.


tshemsedinov:
Refs: #138

belochub:
Landed in https://github.com/metarhia/common/commit/2a04a4946c43d668b8e4cffb4e8fb9aef0626930.

PR #223, opened 2018-11-16
Rename join -> chain and impl Iterator.join similar to Array.join

lundibundi:
* Rename join to chain iterator
This will be a bit more clear and allow to implement 'join' the
similar to the definition in the Array.

* Implement Iterator.join similar to Array.join
It also supports prefix and suffix parameters.

### DIFF
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -159,6 +159,18 @@ class Iterator {
   enumerate() {
     return new EnumerateIterator(this);
   }
+
+  join(sep = ',', prefix = '', suffix = '') {
+    let result = prefix;
+    const { done, value } = this.next();
+    if (!done) {
+      result += value;
+      for (const value of this) {
+        result += sep + value;
+      }
+    }
+    return result + suffix;
+  }
 }
 
 class MapIterator extends Iterator {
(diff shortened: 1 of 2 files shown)


lib/iterator.js
```diff
@@ -159,6 +159,14 @@ class Iterator {
   enumerate() {
     return new EnumerateIterator(this);
   }
+
+  join(sep = ',', prefix = '', suffix = '') {
+    let result = prefix;
+    for (const value of this) {
+      result += value + sep;
+    }
+    return result.slice(0, -sep.length) + suffix;
```
belochub: Nit: Instead of using `slice()` here, I'd prefer to have a separate boolean variable that will determine the addition of the separator, thus avoiding the addition of the separator after the last value. For example:
```js
    let first = true;
    for (const value of this) {
      if (!first) result += sep;
      result += value;
      first = false;
    }
    return result + suffix;
```

nechaido: What about:
```js
let [done, first] = this.next();
if (!done) {
  result += first;
  for (const value of this) {
    resulr += sep + value;
  }  
}
return result + sufix;
```

PR #225, opened 2018-11-23
Allow for Uint64 to be constructed from Int64

belochub:


belochub:
Landed in https://github.com/metarhia/common/commit/c18f7a8aaf89b89ee49040bd3779f1cd89814225.

PR #231, opened 2018-12-04
Implement RangeIterator and Iterator.range() method

belochub:


nechaido:
What do you think about:
```js
function* range(begin, end, step = 1) {
  if (
    (step > 0 && begin >= end) || (step < 0 && begin <= end)
  ) {
    return;
  }
  for (let i = begin; i < end; i += step) {
    yield i;
  }
};

module.exports = {
  Iterator,
  iter,
  range: (begin, end, step = 1) => new Iterator(range(begin, end, step)),
}
```
@belochub @lundibundi

belochub:
@lundibundi @nechaido we can't name the function `range()` since there is already another `range()` function exported from this package placed in `lib/array.js`.

PR #232, opened 2018-12-04
Improve Iterator tests

belochub:
Add missing test cases for Iterator#zip() and Iterator#join().

### DIFF
--- a/test/iterator.js
+++ b/test/iterator.js
@@ -223,7 +223,14 @@ metatests.test('Iterator.flatMap with thisArg', test => {
   test.end();
 });
 
-metatests.test('Iterator.zip', test => {
+metatests.test('Iterator.zip with single iterator', test => {
+  const it = iter(array).take(1);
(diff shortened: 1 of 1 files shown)


test/iterator.js
```diff
@@ -432,3 +439,8 @@ metatests.testSync('Iterator.join with prefix and suffix', test => {
   const actual = iter(array).join(', ', '[', ']');
   test.strictSame(actual, '[1, 2, 3, 4]');
 });
+
+metatests.testSync('Iterator.join on empty iterator', test => {
```
nechaido: I'd prefer to use `test` for consistency.

belochub: All of the other tests on `Iterator.join` use `testSync()`, so it is consistent right now.

belochub:
Landed in https://github.com/metarhia/common/commit/50d34a1d6f2a7402af05d6862fb55fa0fb0ea0f8.

PR #233, opened 2018-12-04
Fix small typo in README

glebavladimir:
Found a typo in your README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Returns: array of objects
 
 Returns: object or array
 
-### Duplicate object ot array
+### Duplicate object or array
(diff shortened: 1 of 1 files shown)


belochub:
Thanks, but this part of the `README.md` is autogenerated from our documentation comments by `metaschema` package, and this typo was fixed there a while ago in 1d25798e99697fbfbd053548194ac2e30f812e0d, we just haven't regenerated the documentation. I think it'd be better to regenerate it after the next `metaschema` version with https://github.com/metarhia/metaschema/pull/107 is published, instead of fixing all of its problems by hand.

Issue #234, opened 2018-12-07
Refactor SID functions to Token


tshemsedinov:
- `crcSID(config, key)` to `crcToken(secret, key)`
- `generateSID(config)` to `generateToken(secret, characters, length)`
- `validateSID(config, sid)` to `validateToken(secret, token)`

PR #236, opened 2018-12-07
Add token validation, deprecate SID functions

tshemsedinov:
Closes: https://github.com/metarhia/common/issues/234

### DIFF
--- a/lib/id.js
+++ b/lib/id.js
@@ -53,6 +53,8 @@ const generateGUID = () => {
 //   config - <Object>, { secret }
 //   key - <string>, SID key
 // Returns: <string>, CRC
+// Deprecated: this method will be removed in the next major versions.
+//     Use `crcToken()` instead.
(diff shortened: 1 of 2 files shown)


lib/id.js
```diff
@@ -73,13 +77,46 @@ const generateSID = config => {
...
+};
+
+// Validate Token
+//   secret - <string>
+//   token - <string>
+// Returns: <boolean>
+const validateToken = (secret, token) => {
+  if (!token) return false;
+  const len = token.length;
+  const crc = token.substr(len - 4);
```
belochub: `String.prototype.substr()` is considered legacy function, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr), I think we can replace it with `String.prototype.slice()` here.

belochub:
Landed in https://github.com/metarhia/common/commit/fb2de22edd749b051edb92fbe7f22f1c983c9e45.

Issue #235, opened 2018-12-07
Name collision in validateHash


tshemsedinov:
First argument `hash` and function `hash`

PR #238, opened 2018-12-07
Fix argument name collision in `validateHash()`

belochub:
Fixes: https://github.com/metarhia/common/issues/235


belochub:
Landed in https://github.com/metarhia/common/commit/837b95ae9f8fd06643004b8af6cb5cc4c250575e.

PR #239, opened 2018-12-10
Fix duplicate() and clone() regression

belochub:
Patch from https://github.com/metarhia/common/pull/85 has introduced a regression in `duplicate()` and `clone()` methods by iterating over Object's properties available from its prototype chain.
This was leading to `TypeError`s in some situations, in particular when cloning Buffer values.

Related CI: https://travis-ci.org/metarhia/metaschema/jobs/466122630


belochub:
Landed in https://github.com/metarhia/common/commit/6d830e56fa1ebca8d84f7e6131744a159526e747.

PR #241, opened 2018-12-11
Version 1.2.1

belochub:
Changelog:

* Fix duplicate() and clone() regression
* Use only obj ownProperties in mergeObjects()


belochub:
Landed in https://github.com/metarhia/common/commit/09070b7916f09663d1f418aa99746edbdd70762b.

PR #243, opened 2019-01-10
Prepare examples of prettier formatted code

JuliaGerasymenko:


.eslintrc.yml
```diff
@@ -1 +1,10 @@
-extends: 'metarhia'
+extends:
+  - metarhia
+  - plugin:prettier/recommended
+  - prettier/react
```
aqrln: You don't need any of these except `metarhia` and `plugin:prettier/recommended`.

test/id.js
```diff
@@ -15,79 +16,107 @@ const config = {
...
-  ],
-  'common.crcSID': [
-    [
-      config,
-      common.generateKey(
-        config.length - 4,
-        config.characters
-      ),
-      result => result.length === 4,
+metatests.case(
```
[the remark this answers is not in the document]
nechaido: I am perfectly fine with proposed changes.

.prettierignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
```
belochub: There is no need to add `node_modules/` to this file; it's ignored by default.

belochub: Also, you still haven't addressed the https://github.com/metarhia/common/pull/243#discussion_r247047923; you must specify the files that should not be formatted by prettier in this file.

README.md
```diff
@@ -15,64 +15,84 @@ $ npm install @metarhia/common
 ```
```
belochub: `README.md` should also be ignored since it is autogenerated.

package.json
```diff
@@ -29,7 +29,8 @@
   ],
   "scripts": {
     "test": "eslint . && metatests test/",
-    "lint": "eslint .",
+    "lint": "eslint . && prettier -c \"**/*.js\" \"**/*.json\" \"**/*.md\" \".*rc\"",
```
belochub: You should also add `*/**.yml` files here and below.

belochub:
Landed in https://github.com/metarhia/common/commit/d018c64c21849de42bb90b9cb243e5b1fb176e0e.

PR #244, opened 2019-01-30
Add simple Pool implementation

belochub:
There are multiple `poolify()` functions implemented in `metasync`
package, but they are quite convoluted, which led to the creation of
this implementation with a simple yet powerful interface having only two
methods and opening up multiple possibilities for optimizations.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -1,0 +1,23 @@
+'use strict';
+
+const storage = Symbol('storage');
+
+class Pool {
+  constructor(factory = null) {
+    this.factory = factory;
+    this[storage] = [];
+  }
+
+  put(value) {
+    this[storage].push(value);
+  }
+
+  get() {
+    if (this[storage].length === 0) {
+      return typeof this.factory === 'function' ? this.factory() : null;
+    }
+    return this[storage].pop();
+  }
+}
+
+module.exports = { Pool };
--- a/test/pool.js
+++ b/test/pool.js
@@ -1,0 +1,43 @@
+'use strict';
+
+const metatests = require('metatests');
+const { Pool } = require('..');
+
+const BUF_LEN = 1000;
+
+metatests.test(
+  'Pool#get() on Pool without factory must return null when empty',
+  test => {
+    const p = new Pool();
+    test.strictSame(p.get(), null);
+    test.end();
+  }
+);
+
+metatests.test(
+  'Pool#get() on Pool without factory must return any value put before',
+  test => {
+    const p = new Pool();
+    for (let i = 0; i < 10; i++) {
+      p.put(Buffer.allocUnsafe(BUF_LEN));
+    }
+    const b = p.get();
+    test.assert(Buffer.isBuffer(b));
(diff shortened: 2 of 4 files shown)


lib/pool.js
```diff
@@ -0,0 +1,27 @@
...
+  constructor(factory) {
+    this.factory = factory;
+    this[storage] = [];
+  }
+
+  put(value) {
+    this[storage].push(value);
+  }
+
+  get() {
```
lundibundi: Perhaps we should allow passing arguments to `factory` here
i.e.
```js
get(...args) {
  ...
  return this.factory(...args);
}
```

test/pool.js
```diff
@@ -0,0 +1,43 @@
...
+    test.assert(Buffer.isBuffer(b));
+    test.strictSame(b.length, BUF_LEN);
+    test.end();
+  }
+);
+
+metatests.test(
+  'Pool#get() on empty Pool with factory must return values retrieved from it',
+  test => {
+    const p = new Pool(() => Buffer.allocUnsafe(BUF_LEN));
```
lundibundi:
```suggestion
    const p = new Pool(test.mustCall(() => Buffer.allocUnsafe(BUF_LEN)));
```
?

lib/pool.js
```diff
@@ -0,0 +1,27 @@
...
+    if (this[storage].length === 0) {
+      if (this.factory) {
+        return this.factory();
+      } else {
+        return null;
+      }
+    }
+    return this[storage].shift();
```
aqrln: I guess using `pop()` would be more efficient?

lib/pool.js
belochub: How would that work? The idea behind the `Pool` is that you always get the same objects, if you need to pass arguments to the factory, you can just bind the factory function before passing it to the constructor or setting the `factory` field of the `Pool` instance to it.
Besides, it would be confusing when you are calling `get()` with arguments, and expect to get an object constructed with those arguments, but instead you get an object that was created before with other arguments and put to the `Pool` instance.

lib/pool.js
SemenchenkoVitaliy: But this will change behaviour - it will be stack instead of the query. Does it matter here?

belochub: @SemenchenkoVitaliy not at all, I can even return a random element here. The single Pool instance is supposed to store multiple instances of the same type and the users of this class are not supposed to depend on the order in which the objects are returned from `get()` after being `put()` into the pool.

lib/pool.js
```diff
@@ -0,0 +1,27 @@
...
+    this[storage].push(value);
+  }
+
+  get() {
+    if (this[storage].length === 0) {
+      if (this.factory) {
+        return this.factory();
+      } else {
+        return null;
+      }
```
tshemsedinov:
```js
      if (this.factory) return this.factory();
      return null;
```

belochub:
@tshemsedinov in my opinion, the asynchronous pool should be implemented in `metasync` package.

PR #249, opened 2019-02-04
Implement recursive mkdir

JuliaGerasymenko:


SemenchenkoVitaliy:
We already have `mkdirp` in [@metarhia/filestorage](https://github.com/metarhia/filestorage/blob/master/lib/utils.js).

/cc @nechaido @belochub @lundibundi @tshemsedinov

nechaido:
@SemenchenkoVitaliy, in that case, we should extract it here.

belochub:
@SemenchenkoVitaliy the purpose of this implementation is to use `fs.mkdir()` with option `recursive` when possible and work as a polyfill in other cases. The implementation in https://github.com/metarhia/filestorage doesn't work like that.

lib/fs.js
```diff
@@ -0,0 +1,28 @@
...
+const mkdir = path => {
+  if (!fs.existsSync(path))
+    fs.mkdirSync(path, err => {
+      if (err) throw err;
+    });
+};
+
+const mkdirp = path => {
+  const version = process.versions.node.split('.').map(el => parseInt(el));
+  if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
```
SemenchenkoVitaliy:
```suggestion
  if (version[0] < 10 || (version[0] >= 10 && version[1] <= 11)) {
```

lib/fs.js
```diff
@@ -0,0 +1,28 @@
...
+    let firstIndex = 0;
+    for (let i = 1; i < path.split('/').length; i++) {
```
SemenchenkoVitaliy: Shouldn't we also support Windows platform?

lib/fs.js
```diff
@@ -0,0 +1,28 @@
...
+      mkdir(path.substring(0, path.indexOf('/', firstIndex)));
+      firstIndex = path.indexOf('/', firstIndex) + 1;
+    }
+    mkdir(path);
+  } else {
+    fs.mkdir(path, { recursive: true }, err => {
+      if (err) throw err;
```
SemenchenkoVitaliy: I'd prefer to return an error to callback instead of throwing it.

lib/fs.js
```diff
@@ -0,0 +1,28 @@
...
+    });
+  }
+};
+
+mkdirp('test/example1/example2');
```
SemenchenkoVitaliy: This should be moved to tests.

lib/fs.js
JuliaGerasymenko: How should I do that? Could you explain in full?

SemenchenkoVitaliy: You can use `path.dirname` to get parent folder and `path.join` to get child folder.

test/fs.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const metatests = require('metatests');
+const common = require('..');
+
+metatests.test('create 2 extra empty directories', test => {
+  common.mkdirp('test/example1/example2');
```
nechaido: You should remove testing dir if it exists, that create it using `mkdirp`, check that it is actually created, and then you should it.

lib/fs.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const fs = require('fs');
+
+const mkdir = path => {
+  if (!fs.existsSync(path))
```
nechaido: AFAIU, we are interested in async `mkdirp`.

lib/fs.js
```diff
@@ -0,0 +1,30 @@
...
+  if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
+    let firstIndex = 0;
+    for (let i = 1; i < path.split('/').length; i++) {
+      mkdir(path.substring(0, path.indexOf('/', firstIndex)));
+      firstIndex = path.indexOf('/', firstIndex) + 1;
+    }
+    mkdir(path);
+  } else {
+    fs.mkdir(path, { recursive: true }, err => {
+      if (err) console.error(err);
```
SemenchenkoVitaliy: As I mentioned in [comment](https://github.com/metarhia/common/pull/249#discussion_r253792011) error should be passed to callback.

lib/fs.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const fs = require('fs');
+
+const mkdir = path => {
+  if (!fs.existsSync(path))
+    fs.mkdirSync(path, err => {
+      if (err) console.error(err);
```
SemenchenkoVitaliy: Same here.

test/fs.js
SemenchenkoVitaliy: You should check if the directory exists and if it does - remove it. Also, it whould be removed after the test anyway.

lib/fs.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+async function mkdirp(pathName, callback) {
```
SemenchenkoVitaliy: We don't need `async/await` here. What @nechaido was talking [here](https://github.com/metarhia/common/pull/249#discussion_r254176051) about is asynchronous function i.e. a function with the callback.

lib/fs.js
```diff
@@ -0,0 +1,26 @@
...
+  const version = process.versions.node.split('.').map(el => parseInt(el));
+  if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
```
SemenchenkoVitaliy: It will be better to move `mkdirp` polyfill to another function since now it will check the version on every iteration.

lib/fs.js
```diff
@@ -0,0 +1,26 @@
...
+    await fs.mkdir(pathName, err => {
```
SemenchenkoVitaliy: `fs.mkdir` does not return `Promise`.

lib/fs.js
```diff
@@ -0,0 +1,26 @@
...
+      if (err) {
+        if (err.code === 'ENOENT') {
+          mkdirp(path.dirname(pathName), callback);
+          mkdirp(pathName, callback);
```
SemenchenkoVitaliy: How is this supposed to work?

test/fs.js
```diff
@@ -0,0 +1,36 @@
...
+const { mkdirp } = require('..');
+const fs = require('fs');
+const path = require('path');
+const pathName = 'test/ex1/ex2';
+
+const callback = msg => {
+  console.log(msg);
+};
+
+function rmdir(path1, callback) {
```
```suggestion
function rmdirp(path1, callback) {
```
?

SemenchenkoVitaliy: Also, it will be better to move it to `lib/fs.js`.

test/fs.js
```diff
@@ -0,0 +1,36 @@
...
+  fs.rmdir(path1, err => {
+    if (err) {
+      if (err.code === 'ENOTEMPTY') {
+        rmdir(path.join(path1, 'ex2'), callback);
+        rmdir(path1, callback);
```
SemenchenkoVitaliy: Same here.

test/fs.js
```diff
@@ -0,0 +1,36 @@
...
+  mkdirp(pathName, callback);
+  setTimeout(() => {
+    fs.stat(pathName, (err, stats) => {
+      if (err) {
+        if (err.code !== 'EEXIST') callback(err);
+      } else if (stats.isDirectory()) {
+        rmdir('test/ex1', callback);
+      }
+    });
+  }, 1000);
```
SemenchenkoVitaliy: You shouldn't use `setTimeout` here.

test/fs.js
```diff
@@ -0,0 +1,36 @@
...
+      if (err.code === 'ENOTEMPTY') {
+        rmdir(path.join(path1, 'ex2'), callback);
+        rmdir(path1, callback);
+      } else if (err.code !== 'EEXIST') callback(err);
+    }
+  });
+}
+
+metatests.test('create 2 extra empty directories', test => {
+  mkdirp(pathName, callback);
```
SemenchenkoVitaliy: The test should be something like this.
```js
rmdirp('test/ex1', err => {
  if (err) {
    test.bailout(err, 'Cannot prepare directory for test');
    return;
  }
  fs.mkdirp(pathName, err => {
    if (err) {
      test.bailout(err, 'Cannot create directory');
      return;
    }
    rmdirp(pathName, err => {
      test.error(err);
      test.end();
    });
  });
});
```

test/fs.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+const metatests = require('metatests');
+const { mkdirp } = require('..');
+const fs = require('fs');
+const path = require('path');
+const pathName = 'test/ex1/ex2';
+
+const callback = msg => {
```
SemenchenkoVitaliy: Since this is the part of tests it should fail tests and not just output the error.

lib/fs.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+  });
+};
+
+const mkdirp = (path, callback) => {
+  const version = process.versions.node.split('.').map(el => parseInt(el));
+  if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
+    mkdir(path, callback);
+  } else {
+    fs.mkdir(path, { recursive: true }, (err, callback) => {
```
SemenchenkoVitaliy:
```suggestion
    fs.mkdir(path, { recursive: true }, callback);
```

lib/fs.js
```diff
@@ -0,0 +1,58 @@
...
+      });
+    } else {
+      cb(err);
+    }
+  });
+};
+
+const mkdirp = (dir, cb) => {
+  const version = process.versions.node.split('.').map(el => parseInt(el));
+  if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
```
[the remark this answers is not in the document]
SemenchenkoVitaliy: Since we might have polyfills in other modules it will be better to extract it to a separate file. Though it should be done in another PR. WDYT?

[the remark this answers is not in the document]
belochub: I don't see any reason for extracting it into a separate file.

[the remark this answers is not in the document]
SemenchenkoVitaliy: For example, at one point we were considering to have `Uint64` and `Int64` based on `BigInt`, though it is not relevant now. Also, some other modules might use it.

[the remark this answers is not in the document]
belochub: And what's that have to do with `fs.mkdir()` polyfill?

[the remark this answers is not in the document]
SemenchenkoVitaliy: Since we will extract function for version compatibility checking, we might as well extract it to a separate file so it can be used by other modules. Though you have the point - this PR shouldn't rely on other PRs.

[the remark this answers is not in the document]
belochub: I haven't proposed to extract the function for version compatibility checking, I've just said that the version check must be performed only once on this file `require()`, not every time the exported function runs since there is no way Node.js version changes between two calls.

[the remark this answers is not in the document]
SemenchenkoVitaliy: Yeah, but since we have version compatibility check in this file, I just thought that we might also implement a separate function for such checks. But, as I said, that's a discussion for a different PR.

common.js
```diff
@@ -11,6 +11,7 @@ const submodules = [
   'events', // Events and emitter
   'flags', // Flags data type
   'fp', // Functional programming
+  'fs', // Recursive mkdir
```
belochub: I think this file will include more file system related methods than just this implementation:
```suggestion
  'fs', // File System
```

test/fs.js
```diff
@@ -0,0 +1,24 @@
...
+
+const metatests = require('metatests');
+const fs = require('fs');
+const { mkdirp } = require('..');
+
+metatests.test('create 2 extra empty directories', test => {
+  mkdirp('test/ex1/ex2', '0777', err => {
+    if (err) {
+      test.bailout(err, 'Cannot create directory');
+      return;
```
belochub: There is no need to use `return` after `test.bailout()` since the latter will throw anyway.

lib/fs.js
```diff
@@ -16,19 +16,22 @@ const mkdir = (dir, mode, cb) => {
...
-
-const mkdirp = (dir, modeValue, cb) => {
-  variant
-    ? fs.mkdir(dir, { recursive: true, mode: modeValue }, cb)
-    : mkdir(dir, modeValue, cb);
+let mkdirp = (dir, modeValue, cb) => {
+  checkVersion(dir, modeValue, cb);
 };
 
+checkVersion = checkVersion();
```
SemenchenkoVitaliy: I'd prefer to avoid unnecessary code:
```js
let mkdirp;
if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
  mkdirp = mkdir;
} else {
  mkdirp = (dir, mode, cb) => fs.mkdir(dir, { recursive: true, mode }, cb);
}
```

SemenchenkoVitaliy: And, as @belochub mentioned in https://github.com/metarhia/common/pull/249#discussion_r256514751 there is no need for a separate `checkVersion` function.

lib/fs.js
```diff
@@ -0,0 +1,30 @@
...
+    }
+  });
+};
+
+let mkdirp;
+const version = process.versions.node.split('.').map(el => parseInt(el));
+if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
+  mkdirp = mkdir;
+} else {
+  mkdirp = (dir, modeValue, cb) =>
```
belochub: The `modeValue` argument must be optional.

lib/fs.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const mkdir = (dir, mode, cb) => {
```
belochub: Same here, the `mode` argument must be optional.

lib/fs.js
```diff
@@ -0,0 +1,34 @@
...
+        else fs.mkdir(dir, cb);
+      });
+    } else {
+      cb(err);
+    }
+  });
+};
+
+const version = process.versions.node.split('.').map(el => parseInt(el));
+const mkdirp = (...args) => {
```
SemenchenkoVitaliy: I don't like rest syntax usage here. I will look much simpler like this
```js
(path, mode, cb) => {
  if (typeof mode === 'function') {
    cb = mode;
    mode = '0777';
  }
```

SemenchenkoVitaliy: Also, you reverted back to checking version every time you call the function which is something we try to avoid.

test/fs.js
```diff
@@ -5,7 +5,7 @@ const fs = require('fs');
 const { mkdirp } = require('..');
 
 metatests.test('create 2 extra empty directories', test => {
-  mkdirp('test/ex1/ex2', err => {
+  mkdirp('test/ex1/ex2', '0777', err => {
```
SemenchenkoVitaliy: It will be nice to have tests with and without mode.

lib/fs.js
```diff
@@ -0,0 +1,43 @@
...
+    mode = '0777';
+  }
+  return [mode, cb];
+};
+
+let mkdirp;
+const version = process.versions.node.split('.').map(el => parseInt(el));
+if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
+  mkdirp = (dir, mode, cb) => {
+    [mode, cb] = addMode(mode, cb);
```
SemenchenkoVitaliy:
```suggestion
  if (typeof mode === 'function') {
    cb = mode;
    mode = '0777';
  }
```
IMO, it will be better to inline this.

lib/fs.js
```diff
@@ -0,0 +1,43 @@
...
+    mkdir(dir, mode, cb);
+  };
+} else {
+  mkdirp = (dir, modeValue, cb) => {
+    [modeValue, cb] = addMode(modeValue, cb);
+    fs.mkdir(dir, { recursive: true, mode: modeValue }, cb);
```
SemenchenkoVitaliy:
```suggestion
    fs.mkdir(dir, { recursive: true, mode }, cb);
```

lib/fs.js
```diff
@@ -0,0 +1,43 @@
...
+    } else {
+      cb(err);
+    }
+  });
+};
+
+const addMode = (mode, cb) => {
+  if (typeof mode === 'function') {
+    cb = mode;
+    mode = '0777';
```
SemenchenkoVitaliy: Put it into some constant at the beginning of the file.
```js
const MKDIRP_DEFAULT_MODE = '0777';
```

SemenchenkoVitaliy:
ping @belochub @nechaido

lib/fs.js
```diff
@@ -0,0 +1,40 @@
...
+    if (typeof mode === 'function') {
+      cb = mode;
+      mode = MKDIRP_DEFAULT_MODE;
+    }
+    mkdir(dir, mode, cb);
+  };
+} else {
+  mkdirp = (dir, mode, cb) => {
+    if (typeof mode === 'function') {
+      fs.mkdir(dir, { recursive: true}, mode);
```
SemenchenkoVitaliy:
```suggestion
      fs.mkdir(dir, { recursive: true }, mode);
```

test/fs.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const metatests = require('metatests');
+const fs = require('fs');
+const { mkdirp } = require('..');
+
+metatests.test('create 2 directories using mode', test => {
```
nechaido: I'd prefer to ensure that directories do not exist prior to `mkdirp`.

test/fs.js
```diff
@@ -0,0 +1,47 @@
...
+  fs.access('test/ex1/ex2', fs.constants.F_OK, err => {
+    if (err) {
```
nechaido:
```suggestion
    if (err && err.code === 'ENOENT') {
```

test/fs.js
```diff
@@ -0,0 +1,47 @@
...
+          if (err) {
+            test.bailout(err, 'Cannot prepare directory for test');
+          }
+          fs.rmdir('test/ex1', err => {
+            test.error(err);
+            test.end();
+          });
+        });
+      });
+    } else test.end();
```
nechaido:
```suggestion
    } else if (err && err.code === 'ENOENT') {
      // Delete directories.
    } else {
      test.bailout(err);
    }
```

SemenchenkoVitaliy:
I suggest using `beforeEach` to remove directories if they exist before tests and `afterEach` to remove them after so tests will have less duplicate code and tests will also look cleaner.

test/fs.js
```diff
@@ -0,0 +1,55 @@
...
+            test.bailout(err, 'Cannot prepare directory for test');
+          }
+          fs.rmdir('test/ex1', err => {
+            test.error(err);
+            test.end();
+          });
+        });
+      });
+    } else if (err && err.code === 'ENOENT') {
+      // Delete directories.
```
nechaido: You should actually delete those directories here.

JuliaGerasymenko:
@SemenchenkoVitaliy I tried to write functions beforeEach and afterEach, removed fs.rmdir inside ones, but it uses test.end, test.bailout, test.error, which makes impossible to remove outside the test. WDYT?

test/fs.js
```diff
@@ -0,0 +1,63 @@
...
+const { mkdirp } = require('..');
+
+metatests.test('create 2 directories using mode', test => {
+  fs.access('test/ex1/ex2', fs.constants.F_OK, err => {
+    if (err === null) {
+      fs.rmdir('test/ex1/ex2', err => {
+        if (err) {
+          test.bailout(err, 'Cannot prepare directory for test');
+        }
+      });
```
```suggestion
        fs.rmdir('test/ex1', err => {
          if (err) test.bailout(err, 'Cannot prepare directory for test');
        });
      });
```

test/fs.js
```diff
@@ -0,0 +1,63 @@
...
+});
+
+metatests.test('create 2 directories without mode', test => {
+  fs.access('test/1/1', fs.constants.F_OK, err => {
+    if (err === null) {
+      fs.rmdir('test/1/1', err => {
+        if (err) {
+          test.bailout(err, 'Cannot prepare directory for test');
+        }
+      });
```
SemenchenkoVitaliy: Same here.

SemenchenkoVitaliy:
@JuliaGerasymenko AFAIK `test.beforeEach` and `test.afterEach` do not call `test.end`, `test.error` or `test.bailout`. Can you please specify what you mean? In `test.beforeEach` you have to check if test directories exist and if they do remove `test/ex1/ex2` and `test/ex1`. The same goes for `test.afterEach` with the only difference is that you don't have to check if they exist.

test/fs.js
```diff
@@ -0,0 +1,56 @@
...
+    if (err) console.log(err);
+    fs.rmdir(path.dirname(dir), err => {
+      if (err) console.log(err);
+    });
+  });
+};
+
+metatests.test('create 2 directories using mode', test => {
+  const dir1 = 'test/ex1/ex2';
+  beforeEach(dir1);
```
SemenchenkoVitaliy: Same here: there will be a race and it can happen that directories will be removed by `beforeEach` after `mkdirp`.

test/fs.js
```diff
@@ -0,0 +1,56 @@
...
+  fs.access(dir1, fs.constants.F_OK, err => {
+    if (err === null || (err && err.code === 'ENOENT')) {
+      mkdirp(dir1, 0o777, err => {
+        if (err) test.bailout(err, 'Cannot create directory');
+        afterEach(dir1);
+      });
+    }
+    test.end();
```
SemenchenkoVitaliy: The test will end even before `mkdirp` is called.

test/fs.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+const metatests = require('metatests');
+const fs = require('fs');
+const { mkdirp } = require('..');
+const path = require('path');
+
+const beforeEach = dir => {
```
SemenchenkoVitaliy: It should be `test.beforeEach` and `test.afterEach`. They will be called before and after each test, so you don't need different directories for each test like 'test/ex1/ex2' and 'test/1/2', it will be only one of them.

test/fs.js
```diff
@@ -0,0 +1,56 @@
...
+      fs.rmdir(path.dirname(dir), err => {
+        if (err) console.log(err);
+      });
+    }
+  });
+};
+
+const afterEach = dir => {
+  fs.rmdir(dir, err => {
+    if (err) console.log(err);
```
SemenchenkoVitaliy: You cannot just log errors since it won't fail tests if there is any error.

test/fs.js
```diff
@@ -0,0 +1,56 @@
...
+      if (err) console.log(err);
+    });
+  });
+};
+
+metatests.test('create 2 directories using mode', test => {
+  const dir1 = 'test/ex1/ex2';
+  beforeEach(dir1);
+  fs.access(dir1, fs.constants.F_OK, err => {
+    if (err === null || (err && err.code === 'ENOENT')) {
```
SemenchenkoVitaliy: `err === null`  means that `dir1` already exists which means that `mkdirp` won't do anything.

test/fs.js
```diff
@@ -0,0 +1,56 @@
...
+      mkdirp(dir1, 0o777, err => {
+        if (err) test.bailout(err, 'Cannot create directory');
+        afterEach(dir1);
+      });
+    }
+    test.end();
+  });
+});
+
+metatests.test('create 2 directories without mode', test => {
```
SemenchenkoVitaliy: Comments for the previous test also apply here.

test/fs.js
```diff
@@ -0,0 +1,49 @@
...
+    fs.rmdir(path.dirname(mkdirpTestDir), err => {
+      if (err) test.error(err);
+    });
+    cb();
+  });
+});
+
+testMkdirp.test('create 2 directories using mode', test => {
+  mkdirp(mkdirpTestDir, 0o777, err => {
+    if (err) test.bailout(err, 'Cannot create directory');
```
SemenchenkoVitaliy: I think `test.error` should be enough.
```suggestion
    test.error(err, 'Cannot create directory');
```

test/fs.js
```diff
@@ -0,0 +1,49 @@
...
+    test.end();
+  });
+});
+
+testMkdirp.test('create 2 directories without mode', test => {
+  mkdirp(mkdirpTestDir, err => {
+    if (err) test.bailout(err, 'Cannot create directory');
```
SemenchenkoVitaliy: Same here.
```suggestion
    test.error(err, 'Cannot create directory');
```

test/fs.js
```diff
@@ -0,0 +1,49 @@
...
+        test.error(err);
+      }
+      cb();
+    });
+  });
+});
+
+testMkdirp.afterEach((test, cb) => {
+  fs.rmdir(mkdirpTestDir, err => {
+    if (err) test.error(err);
```
```suggestion
    if (err && err.code !== 'ENOENT') test.error(err);
```

test/fs.js
```diff
@@ -0,0 +1,49 @@
...
+    cb();
+    fs.rmdir(path.dirname(mkdirpTestDir), err => {
+      if (err) test.error(err);
```
SemenchenkoVitaliy: Same here and `cb()` call should be here.
```suggestion
      if (err && err.code !== 'ENOENT') test.error(err);
      cb();
```

test/fs.js
```diff
@@ -0,0 +1,49 @@
...
+      }
+      cb();
+    });
+  });
+});
+
+testMkdirp.afterEach((test, cb) => {
+  fs.rmdir(mkdirpTestDir, err => {
+    if (err) test.error(err);
+    cb();
```
SemenchenkoVitaliy: This `cb()` call is not needed here.
```suggestion
```

test/fs.js
```diff
@@ -0,0 +1,49 @@
...
+    fs.rmdir(path.dirname(mkdirpTestDir), err => {
+      if (err) test.error(err);
+    });
+    cb();
```
SemenchenkoVitaliy: This callback should be inside callback of second `fs.rmdir`.
```suggestion
```

test/fs.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+const metatests = require('metatests');
+const fs = require('fs');
+const { mkdirp } = require('..');
+const path = require('path');
```
SemenchenkoVitaliy: Please, add an empty line after this.
```suggestion
const path = require('path');

```

test/fs.js
```diff
@@ -0,0 +1,49 @@
...
+const testMkdirp = metatests.test('mkdir');
+const mkdirpTestDir = 'test/ex1/ex2';
+
+testMkdirp.beforeEach((test, cb) => {
```
SemenchenkoVitaliy: `testMkdirp.beforeEach` should be identical to `testMkdirp.afterEach`.

test/fs.js
```diff
@@ -0,0 +1,47 @@
...
+const fs = require('fs');
+const { mkdirp } = require('..');
+const path = require('path');
+
+const testMkdirp = metatests.test('mkdir');
+const mkdirpTestDir = 'test/ex1/ex2';
+
+testMkdirp.beforeEach((test, cb) => {
+  fs.rmdir(mkdirpTestDir, err => {
+    if (err && err.code === 'ENOENT') {
```
SemenchenkoVitaliy: Should be
```suggestion
    if (err && err.code !== 'ENOENT') test.error(err);
```
just like in `testMkdirp.afterEach`.

SemenchenkoVitaliy:
ping @belochub.

test/fs.js
```diff
@@ -0,0 +1,43 @@
...
+    if (err && err.code !== 'ENOENT') test.error(err);
+    fs.rmdir(path.dirname(mkdirpTestDir), err => {
+      if (err && err.code !== 'ENOENT') test.error(err);
+      cb();
+    });
+  });
+});
+
+testMkdirp.afterEach((test, cb) => {
+  fs.rmdir(mkdirpTestDir, err => {
```
lundibundi: Why not extract this into a function and then use
```js
const removeUserDirs = (test, cb) => {};
testMkdirp.beforeEach(removeUserDirs);
testMkdirp.afterEach(removeUserDirs);
```
?

PR #251, opened 2019-02-12
Recursive dir copy

KHYehor:
Implementation of recursive dir copy with several options but without deleting the source dir while coping and tests
Don't you think that it is better to do as another function?

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -62,7 +62,103 @@ if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
   };
 }
 
+const metasync = require('metasync');
+
+const error = (err, options, cb) => {
+  if (options.errorLog) options.errorLog.write(err.stack);
+  cb(err);
+};
+
+const cpDir = (src, dest, options, curDepth, cb) => {
+  if (curDepth > options.depth) {
+    cb();
+    return;
+  }
+  fs.mkdir(dest, err => {
+    if (err !== null && err.code !== 'EEXIST') {
+      error(err, options, cb);
+      return;
+    }
+    fs.readdir(src, (err, files) => {
+      if (err) {
+        error(err, options, cb);
+        return;
+      }
+      metasync.each(
+        files,
+        (file, cb) => {
+          if (
+            (options.filter instanceof RegExp && options.filter.test(file)) ||
+            (typeof options.filter === 'function' && options.filter(file))
+          ) {
+            cb();
+            return;
+          }
+          const srcFile = path.join(src, file);
+          fs.lstat(srcFile, (err, stats) => {
+            if (err) {
+              error(err, options, cb);
+              return;
+            }
+            const destFile = path.join(dest, file);
+            if (stats.isDirectory()) {
+              cpDir(srcFile, destFile, options, curDepth + 1, cb);
+            } else if (stats.isFile()) {
+              if (options.Transform) {
+                const read = fs.createReadStream(srcFile, options.readOptions);
+                const write = fs.createWriteStream(
+                  destFile,
+                  options.writeOptions
+                );
+                read.pipe(new options.Transform()).pipe(write);
+                read.on('error', cb);
+                write.on('error', cb).on('close', cb);
+              } else {
+                fs.copyFile(srcFile, destFile, cb);
+              }
+            } else if (stats.isSymbolicLink()) {
+              if (options.dereference) {
+                fs.readlink(srcFile, (err, link) => {
+                  if (err) {
+                    error(err, options, cb);
+                    return;
+                  }
+                  cpDir(
+                    path.resolve(path.dirname(srcFile), link),
+                    destFile,
+                    options,
+                    curDepth + 1,
+                    cb
+                  );
+                });
+              } else {
+                fs.readlink(srcFile, (err, link) => {
+                  if (err) {
+                    error(err, options, cb);
+                    return;
+                  }
+                  fs.symlink(link, destFile, cb);
+                });
+              }
+            }
+          });
+        },
+        cb
+      );
+    });
+  });
+};
+
+const copyDir = (src, dest, options, cb) => {
+  if (typeof options === 'function') {
+    cb = options;
+    options = {};
+  }
+  cpDir(src, dest, options, 1, cb);
+};
+
 module.exports = {
+  copyDir,
   mkdirp,
   rmdirp,
 };
--- a/test/fs.js
+++ b/test/fs.js
@@ -2,12 +2,15 @@
 
 const metatests = require('metatests');
 const fs = require('fs');
+const metasync = require('metasync');
+const common = require('..');
+const stream = require('stream');
 const path = require('path');
 const { mkdirp, rmdirp } = require('..');
 
 const testMkdirp = metatests.test('mkdir');
 const mkdirpTestDir = 'test/ex1/ex2';
-const RMDIRP_TEST_DIR = 'testDir';
+const RMDIRP_TEST_DIR = 'tstDir';
 
 const removeUsedDirs = (test, cb) => {
   fs.rmdir(mkdirpTestDir, err => {
@@ -57,3 +60,251 @@ metatests.test('rmdirp test', test =>
     });
   })
 );
+
+const testData =
+  'JySAF1figbONWxAeeoGhDXeI7df25GKXKPcyCl63ZjIv5HZGQH8yfFeHGGaSJ9zDMP0CwtgX2madIQkFmuq38SNhcKcGQ';
+const absoluteLink = __dirname + '/lDir';
+const srcFiles = [
+  'test/srcDir',
+  'test/srcDir/dir1',
+  'test/srcDir/dir2',
+  'test/srcDir/file1',
+  'test/srcDir/link',
+  'test/srcDir/dir1/subdir1',
+  'test/lDir',
+  'test/lDir/file1',
+  'test/lDir/file2',
+];
+
+const dstFiles = [
+  'test/dstDir',
+  'test/dstDir/dir1',
+  'test/dstDir/dir2',
+  'test/dstDir/file1',
+  'test/dstDir/link',
+  'test/dstDir/dir1/subdir1',
+];
+
+class ToUpperCase extends stream.Transform {
+  constructor(options = {}) {
+    options.decodeStrings = false;
+    super(options);
+  }
+  _transform(chunk, encoding, callback) {
+    if (encoding === 'utf8') callback(null, chunk.toUpperCase());
+    else callback(new Error('Only utf8 is supported'));
+  }
+}
+
+const transOptions = {
+  Transform: ToUpperCase,
+  readOptions: { encoding: 'utf-8' },
+  writeOptions: {},
+};
+
+const rmdir = (dir, cb) => {
+  fs.lstat(dir, (err, stats) => {
+    if (err) {
+      if (err.code === 'ENOENT') cb(null);
+      else cb(err);
+      return;
+    }
+    if (stats.isDirectory()) {
+      fs.readdir(dir, (err, files) => {
+        if (err) {
+          cb(err);
+          return;
+        }
+        files = files.map(file => path.join(dir, file));
+        metasync.each(files, rmdir, err => {
+          if (err) cb(err);
+          else fs.rmdir(dir, cb);
+        });
+      });
+    } else {
+      fs.unlink(dir, cb);
+    }
+  });
+};
+
+const copydirTest = metatests.test('Recursive dir copy');
+copydirTest.endAfterSubtests();
+
+copydirTest.beforeEach((test, cb) => {
+  metasync.series(
+    srcFiles,
+    (file, cb) => {
+      if (path.basename(file).startsWith('file')) {
+        fs.appendFile(file, testData, cb);
+      } else if (path.basename(file).startsWith('link')) {
+        fs.symlink(absoluteLink, file, cb);
+      } else {
+        fs.mkdir(file, cb);
+      }
+    },
+    test.cb(cb)
+  );
+});
+
+copydirTest.afterEach((test, cb) => {
+  rmdir(dstFiles[0], err => {
+    if (err) test.error(err);
+    rmdir(srcFiles[0], err => {
+      if (err) test.error(err);
+      rmdir('test/lDir', test.cb(cb));
+    });
+  });
+});
+
+copydirTest.test('copy', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(
+      dstFiles,
+      (file, cb) => {
+        const fileName = path.basename(file);
+        if (fileName.startsWith('file')) {
+          fs.readFile(file, 'utf-8', (err, data) => {
+            test.strictSame(data, testData);
+            cb(err);
+          });
+        } else {
+          fs.access(file, cb);
+        }
+      },
+      test.cb(() => test.end())
+    );
+  });
+});
+
+copydirTest.test('dereference', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], { dereference: true }, err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(
+      common.merge(dstFiles, [
+        'test/dstDir/link/file1',
+        'test/dstDir/link/file2',
+      ]),
+      (file, cb) => {
+        const fileName = path.basename(file);
+        if (fileName.startsWith('file')) {
+          fs.readFile(file, 'utf-8', (err, data) => {
+            test.strictSame(data, testData);
+            cb(err);
+          });
+        } else {
+          fs.access(file, cb);
+        }
+      },
+      test.cb(() => test.end())
+    );
+  });
+});
+
+copydirTest.test('transform', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], transOptions, err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(
+      dstFiles,
+      (file, cb) => {
+        const fileName = path.basename(file);
+        if (fileName.startsWith('file')) {
+          fs.readFile(file, 'utf-8', (err, data) => {
+            test.strictSame(data, testData.toUpperCase());
+            cb(err);
+          });
+        } else {
+          fs.access(file, cb);
+        }
+      },
+      test.cb(() => test.end())
+    );
+  });
+});
+
+copydirTest.test('filter', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], { filter: /file/ }, err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(
+      common.merge(dstFiles.slice(0, 3), dstFiles.slice(-2)),
+      (file, cb) => {
+        const fileName = path.basename(file);
+        if (fileName.startsWith('file')) {
+          fs.readFile(file, 'utf-8', (err, data) => {
+            test.strictSame(data, testData);
+            cb(err);
+          });
+        } else {
+          fs.access(file, cb);
+        }
+      },
+      test.cb(() => test.end())
+    );
+  });
+});
+
+copydirTest.test('depth more than expected', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], { depth: 100 }, err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(
+      dstFiles,
+      (file, cb) => {
+        const fileName = path.basename(file);
+        if (fileName.startsWith('file')) {
+          fs.readFile(file, 'utf-8', (err, data) => {
+            test.strictSame(data, testData);
+            cb(err);
+          });
+        } else {
+          fs.access(file, cb);
+        }
+      },
+      test.cb(() => test.end())
+    );
+  });
+});
+
+copydirTest.test('depth less than expected', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], { depth: 2 }, err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(
+      dstFiles.slice(0, 5),
+      (file, cb) => {
+        const fileName = path.basename(file);
+        if (fileName.startsWith('file')) {
+          fs.readFile(file, 'utf-8', (err, data) => {
+            test.strictSame(data, testData);
+            cb(err);
+          });
+        } else {
+          fs.access(file, cb);
+        }
+      },
+      test.cb(() => test.end())
+    );
+  });
+});
--- a/common.js
+++ b/common.js
@@ -20,6 +20,7 @@ const submodules = [
   'network', // Network utilities
   'oop', // Object-oriented programming
   'pool', // Object pool
+  'fs', // file system utilities
   'sort', // Sort compare functions
   'strings', // Strings utilities
   'time', // Date and Time functions


lib/rdc.js
```diff
@@ -0,0 +1,97 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const rdc = (source, destination, options, callback) => {
+  if (typeof callback !== 'function') {
```
SemenchenkoVitaliy:
```js
  if (typeof options === 'function') {
     callback = options;
     options = {};
   }
```

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+    if (typeof options !== 'function') throw new Error('ERR_INVALID_CALLBACK');
+    callback = options;
+    options = {};
+  }
+
+  const depth = options.depth;
```
SemenchenkoVitaliy:
```js
const { depth, filter, stopOnError: stop, errorLog, dereference} = options;
```

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+  const Transform = options.change.transform;
```
SemenchenkoVitaliy:
```js
const { transform, readStreamOptions, writeStreamOptions } = options.change;
```

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const rdc = (source, destination, options, callback) => {
```
SemenchenkoVitaliy:
```suggestion
const copyDir = (source, destination, options, callback) => {
```

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+  const dereference = options.dereference;
+
+  const error = err => {
+    if (stop) throw new Error(err);
+    else if (errorLog) errorLog.write(JSON.stringify(err));
+    else callback(err);
+  };
+
+  const copyDirectory = target => {
+    fs.mkdirSync(target, err => {
```
SemenchenkoVitaliy: It should be `fs.mkdir`.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+              copyFile(current, target);
+            }
+          }
+        });
+      });
+    });
+  };
+
+  if (!fs.existsSync(destination)) {
+    fs.mkdirSync(destination);
```
SemenchenkoVitaliy: ditto.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+            } else if (typeof filter === 'undefined') {
+              copyFile(current, target);
+            }
+          }
+        });
+      });
+    });
+  };
+
+  if (!fs.existsSync(destination)) {
```
SemenchenkoVitaliy: ditto. It's better to avoid synchronous functions.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  const copyFile = (current, target) => {
+    if (Transform) {
+      fs.createReadStream(current, readStreamOptions)
+        .pipe(new Transform())
+        .pipe(fs.createWriteStream(target, writeStreamOptions));
+    } else {
+      fs.copyFile(current, target, err => {
+        if (err) error(err);
```
SemenchenkoVitaliy: If the error happened you should return it to the callback.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+  const Transform = options.change.transform;
+  const readStreamOptions = options.change.readStreamOptions;
+  const writeStreamOptions = options.change.writeStreamOptions;
+  const filter = options.filter;
+  const stop = options.stopOnError;
+  const errorLog = options.errorLog;
+  const dereference = options.dereference;
+
+  const error = err => {
+    if (stop) throw new Error(err);
```
SemenchenkoVitaliy: I'd prefer to pass error to the callback, not throw it.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+    else if (errorLog) errorLog.write(JSON.stringify(err));
+    else callback(err);
+  };
+
+  const copyDirectory = target => {
```
SemenchenkoVitaliy: It should be moved outside function.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+    fs.mkdirSync(target, err => {
+      if (err) error(err);
+    });
+  };
+
+  const copyFile = (current, target) => {
```
SemenchenkoVitaliy: Same here.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+        const target = path.resolve(dest, item);
+        fs.lstat(current, (err, stat) => {
+          if (err) error(err);
+          else if (stat.isDirectory()) {
+            copyDirectory(target);
+            start(current, target, ++curDepth);
+          } else if (stat.isSymbolicLink()) {
+            copyLink(current, target);
+          } else if (stat.isFile()) {
+            if (filter instanceof RegExp && !filter.test(item)) {
```
SemenchenkoVitaliy: What is the point of branching if you make the same action?

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+        fs.symlink(linkString, target, err => {
+          if (err) error(err);
+        });
+      }
+    });
+  };
+
+  const start = (curDir = source, dest = destination, curDepth = 1) => {
+    if (curDepth > depth) return;
+    fs.readdir(curDir, (err, items) => {
```
SemenchenkoVitaliy: I don't see where you call the callback when you succeeded in copying.

lib/rdc.js
```diff
@@ -0,0 +1,97 @@
...
+      if (err) error(err);
+      items.forEach(item => {
```
SemenchenkoVitaliy: It will be better to use [@metarhia/metasync](github.com/metarhia/metasync) for such kind of array processing.

SemenchenkoVitaliy:
Also, look at CI build and fix linter issues.

lib/rdc.js
KHYehor: I can't move this function outside  main because I use ```error()``` fn
```suggestion
  const copyDirectory = target => {
```

lib/rdc.js
KHYehor: the same here

lib/rdc.js
SemenchenkoVitaliy: Then why not change signature of `copyDirectory` to something like this?
```js
  const copyDirectory = (target, options, callback) => {
```
First of all, this shouldn't be named `copyDirectory` since it basically calls `fs.mkdirSync`. Also, it should be asynchronous. And at that point why not just inline it?

lib/rdc.js
KHYehor: I don't understand what do you mean, should I use metasync for ```fs.readdir``` or ```for items.forEach``` or for both?

lib/rdc.js
KHYehor: Yes, it is my mistake. What do you think about using counter to check the last file coping and then call callback(null)? or think about another solution?

SemenchenkoVitaliy: That is related to this [comment](https://github.com/metarhia/common/pull/251/files#r256290296).
This is example of basic recursive directory copying.
```js
const copydir = (src, dest, cb) => {
  fs.readdir(dest, (err, files) => {
    if (err) {
      cb(err);
      return;
    }
    metasync.each(files, (file, cb) => {
      fs.stat(file, (err, stats) => {
        if (err) {
          cb(err);
          return;
        }
        const srcFile = path.join(src, file);
        const destFile = path.join(dest, file);
        if (stats.isDirectory()) {
          fs.mkdir(destFile, err => {
            if (err) cb(err);
            else copydir(srcFile, destFile, cb);
          });
        } else {
          fs.copyFile(srcFile, destFile, cb);
        }
      });
    }, cb);
  });
};
```

lib/fs.js
```diff
@@ -0,0 +1,129 @@
...
+const copyFile = promisify(fs.copyFile);
+const symlink = promisify(fs.symlink);
+const readlink = promisify(fs.readlink);
+const lstat = promisify(fs.lstat);
+const readdir = promisify(fs.readdir);
+const mkdir = promisify(fs.mkdir);
+const rmdir = promisify(fs.rmdir);
+const unlink = promisify(fs.unlink);
+
+const rmDir = (current, callback) => {
```
SemenchenkoVitaliy: There is a separate PR for that https://github.com/metarhia/common/pull/252

package-lock.json
```diff
@@ -972,21 +972,10 @@
...
-      },
-      "dependencies": {
-        "acorn": {
-          "version": "5.7.3",
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.7.3.tgz",
-          "integrity": "sha512-T/zvzYRfbVojPWahDsE5evJdHb3oJoQfFbsrKM7w5Zcs++Tr257tia3BmMP8XYVjp1S9RZXQMh7gao96BlqZOw==",
-          "dev": true
-        }
-      }
+      "version": "4.0.0",
```
SemenchenkoVitaliy: We usually only update `package-lock.json` if PR changes dependencies or on a new version release.

lib/fs.js
```diff
@@ -0,0 +1,129 @@
...
+  const error = (err, resolve, reject) => {
+    if (err.code === 'EEXIST') resolve();
+    else if (!errorLog) reject(err);
+    else {
+      errorLog.write(JSON.stringify(err));
+      resolve();
+    }
+  };
+
+  const cpFile = (current, target) => {
```
SemenchenkoVitaliy: This should be moved outside the function or inlined as you use it only once. Also, in this case, it is better to pass variables you need as variables then rely on context.

lib/fs.js
```diff
@@ -0,0 +1,129 @@
...
+        const read = fs.createReadStream(current, readOptions);
+        const write = fs.createWriteStream(target, writeOptions);
+        read.pipe(new Transform()).pipe(write);
+        read.on('error', reject);
+        write.on('error', reject).on('close', () => resolve());
+      });
+    else return copyFile(current, target);
+  };
+
+  const cpItem = (current, target, curDepth) =>
```
SemenchenkoVitaliy: Same here.

lib/fs.js
```diff
@@ -0,0 +1,129 @@
...
+            else return new Promise(resolve => resolve());
+          } else if (stat.isFIFO()) {
+            return symlink(current, target);
+          }
+        })
+        .then(() => resolve())
+        .catch(err => error(err, resolve, reject))
+    );
+
+  const cpDir = (source, destination, curDepth = 1) =>
```
SemenchenkoVitaliy: ditto.

SemenchenkoVitaliy:
https://github.com/metarhia/common/pull/251#issuecomment-463113845 Still relevant.

lib/fs.js
KHYehor: I realised, that I can't catch moment when the last file is copied to call callback(null) at the end, so I was suggested to use promises.

SemenchenkoVitaliy:
Response to https://github.com/metarhia/common/pull/251#discussion_r257460532:

> I realised, that I can't catch moment when the last file is copied to call callback(null) at the end, so I was suggested to use promises.

As I previously mentioned [`metasync.each` can be used](https://github.com/metarhia/common/pull/251#discussion_r256560678) for this. If you have problems with its usage take a look at [function](https://github.com/metarhia/metasync/blob/master/lib/array.js) and [usage examples](https://github.com/metarhia/metasync/blob/master/test/array.each.js) or ask a question.

lib/fs.js
SemenchenkoVitaliy: If you answer on the review summary don't post it on unrelated review comment.

KHYehor:
Ok, as said Travis, I need to rewrite my promisify tests, So it will be better to replace it with metasync again, won't it?

KHYehor:
Can you please explain me, what should I do if Travis says me ```Failure after end: fs.copyFile is not a function```?

tshemsedinov:
@KHYehor try to run `npm t` on your pc before commit

KHYehor:
@tshemsedinov Yes, I have done it, ```Metatests finished with code 0```
Travis says that there isn't ```fs.copyFile``` fn for v6.16.0, I checked documentation and it is true.

SemenchenkoVitaliy:
@KHYehor `fs.copyFile` [was added in `v8.5.0`](https://nodejs.org/api/fs.html#fs_fs_copyfile_src_dest_flags_callback). Where did you get information about `v6.16.0`?

KHYehor:
@SemenchenkoVitaliy I get it right there https://travis-ci.org/metarhia/common/jobs/494536979

tshemsedinov:
We can drop 6.x support

KHYehor:
Ok) So, what have I do? Can I turn off Travis for checking v6.x?

SemenchenkoVitaliy:
@KHYehor You can wait until https://github.com/metarhia/common/pull/255 is landed and then rebase on master. Until then only check if CI for Node.js 8, 10, 11 is not failed.

KHYehor:
@SemenchenkoVitaliy I have one question, Should I revert ```package-lock.json``` to older version? As I realised I failed with it updating, didn't I?

SemenchenkoVitaliy:
@KHYehor Yes, since this PR does not change dependencies there is no need to update `package-lock.json` as I mentioned [here](https://github.com/metarhia/common/pull/251#discussion_r257459314).

KHYehor:
I replaced ```package-lock.json``` from master branch.

package-lock.json
```diff
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@babel/cli": {
-      "version": "7.1.5",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.1.5.tgz",
-      "integrity": "sha512-zbO/DtTnaDappBflIU3zYEgATLToRDmW5uN/EGH1GXaes7ydfjqmAoK++xmJIA+8HfDw7UyPZNdM8fhGhfmMhw==",
+      "version": "7.2.3",
```
SemenchenkoVitaliy: Changes in `package-lock.json` should be reverted to your local `master` branch using `git reset master package-lock.json`. Or you can make `git fetch` and rebase onto `origin/master`.

lib/fs.js
```diff
@@ -0,0 +1,93 @@
...
+          ) {
+            fs.stat(path.join(src, file), (err, stats) => {
+              if (err) {
+                error(err, options, cb);
+                return;
+              }
+              const srcFile = path.join(src, file);
+              const destFile = path.join(dest, file);
+              if (stats.isDirectory())
+                cpDir(srcFile, destFile, options, cb, ++curDepth);
```
SemenchenkoVitaliy:
```suggestion
                cpDir(srcFile, destFile, options, cb, curDepth + 1);
```
If you have more than one nested directory on the same level it will increment `curDepth` for each directory, which will lead to incorrect behaviour.

lib/fs.js
```diff
@@ -0,0 +1,93 @@
...
+        error(err, options, cb);
+        return;
+      }
+      metasync.each(
+        files,
+        (file, cb) => {
+          if (
+            (options.filter instanceof RegExp && !options.filter.test(file)) ||
+            (typeof options.filter === 'function' && !options.filter(file)) ||
+            typeof options.filter === 'undefined'
```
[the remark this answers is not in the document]
KHYehor: So, just remove this if case?
or what do you mean?

lib/fs.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+    fs.readdir(src, (err, files) => {
+      if (err) {
+        error(err, options, cb);
+        return;
+      }
+      metasync.each(
+        files,
+        (file, cb) => {
+          if (
```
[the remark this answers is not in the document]
KHYehor: Ok, I have understood.

lib/fs.js
```diff
@@ -0,0 +1,95 @@
...
+            cb();
+            return;
+          }
+          const srcFile = path.join(src, file);
+          fs.stat(srcFile, (err, stats) => {
+            if (err) {
+              error(err, options, cb);
+              return;
+            }
+            const srcFile = path.join(src, file);
```
SemenchenkoVitaliy:
```suggestion
```
You declared it previously, no need to have a duplicate

lib/fs.js
```diff
@@ -0,0 +1,95 @@
...
+            const destFile = path.join(dest, file);
+            if (stats.isDirectory())
```
SemenchenkoVitaliy:
```js
            if (stats.isDirectory()) {
              cpDir(srcFile, destFile, options, curDepth + 1, cb);
            } else if (stats.isFile()) {
```

test/fs.js
```diff
@@ -0,0 +1,186 @@
...
+      this.emit('error', new Error('only utf8'));
+      return callback();
+    }
+    this.push(chunk.toUpperCase());
+    return callback();
+  }
+}
+
+metatests.test('fs / copyDir', test => {
+  createTestDir(srcDir + '1', err => {
```
SemenchenkoVitaliy: It will be better to use `beforeEach` and `afterEach` to avoid boilerplate code.

test/fs.js
```diff
@@ -0,0 +1,186 @@
...
+  '3.html',
+  '4.css',
+  '5.doc',
+  '6.tmp',
+  '7.asm',
+  '8.zip',
+  'link',
+  'dir1',
+  'dir2',
+  'dir3',
```
SemenchenkoVitaliy: Why did you reduce the number of nested directories?

test/fs.js
```diff
@@ -0,0 +1,186 @@
...
+        });
+      }
+    );
+  });
+});
+
+metatests.test('fs / copyDir / depth', test => {
+  createTestDir(srcDir + '5', err => {
+    if (err) test.error(err);
+    copyDir(srcDir + '5', targetDir + '5', { depth: 2 }, err => {
```
SemenchenkoVitaliy: There also should be a test which calls `copyDir` with `depth` which is higher than levels of nested files and vise versa. Also, you should check if all files that are supposed to be copied are indeed were copied and files that should not be copied are not copied.

lib/fs.js
KHYehor: What is wrong? It is the same, or I just don't see something?

test/fs.js
KHYehor: I can change it back, my OS had limit for opened files at once.

lib/fs.js
SemenchenkoVitaliy: Yes, formatting. If the code in `if` is too long to put it on one line or if there is `else` it should be wrapped in curly braces just like you did [here](https://github.com/metarhia/common/pull/251/files#diff-9a205ef7ee967ee32efee02e58b3482dR61). It is not obligatory, it doesn't impact performance, but it improves readability and makes easier to make changes in the future. You can check out code style recommendations [here](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/conditionals).

test/fs.js
SemenchenkoVitaliy: Well, it should be fine though I wonder how tests passed before that?

test/fs.js
KHYehor: About the last sentence, So, just do something like recursive ```fs.exist```?
Can I change timeout limit somehow? Because if I start 6 tests instead of 5 I get timeout exception, Seems like It needs just some more time to finish deleting, but I am not sure

SemenchenkoVitaliy: `fs.exists` is deprecated, so it will be better to use `fs.access` instead. Also, do you get a timeout for all tests or just for a particular one? I added several tests and everything seems fine. FYI, timeout is 30 seconds.

test/fs.js
```diff
@@ -0,0 +1,186 @@
...
+      return callback();
+    }
+    this.push(chunk.toUpperCase());
+    return callback();
+  }
+}
+
+metatests.test('fs / copyDir', test => {
+  createTestDir(srcDir + '1', err => {
+    if (err) test.error(err);
```
```suggestion
    if (err) {
      test.error(err);
      return;
    }
```
If test cannot be continued without previous operation being successful, there should be `return`.

SemenchenkoVitaliy: Also, there is now `test.cb` and `test.cbFail` in [metatests 0.5.0](https://github.com/metarhia/metatests) which can be used to reduce manual error handling.

test/fs.js
KHYehor: Again question about ```beforeEach``` and ```afterEach```, are these something from metatests?

test/fs.js
KHYehor: What do you think about getting array of files from ```fs.readdir``` and then compare it with static array files?

KHYehor: I have changed default timeout to 45000 and all tests are passed
Seems like it is hard for my OS to create-copy-delete parallel 6 dirs
also my cooler rotates faster when I check my tests, I hope I won't kill my laptop

test/fs.js
SemenchenkoVitaliy: Yes, you can learn about it more [here](https://github.com/metarhia/metatests/blob/d5fbe7827bd388d9d78c2bb7f98cd02b83331562/lib/imperative-test.js) and [here](https://github.com/metarhia/filestorage/blob/master/test/filestorage.js). It will allow executing the function before and after the specified set of tests e.i. create test files and clean up them.

test/fs.js
SemenchenkoVitaliy: > Seems like it is hard for my OS to create-copy-delete parallel 6 dirs

That is ok, as long as CI passes there is no need to do something about this.

> What do you think about getting array of files from `fs.readdir` and then compare it with static array files?

Yeah, this is fine. BTW, there is no need in so many files. I think 2 files for each level should be enough.

KHYehor: Also one question from me: I need fn which compare two arrays for equality
Does Metarhia have something like this fn, or I should do it in my own?

SemenchenkoVitaliy: You need to sort `files` from `fs.readdir` and then compare using `test.strictSame`.
```js
test.strictSame(files.sort(), expected)
```

KHYehor: And as I understood I should give ```test``` as argument to another fn which checks for equality
something like this
``` javascript
const testTestDir = (current, test, cb) => {
  fs.readdir(current, (err, items) => {
    if (err) {
      cb(err);
      return;
    }
    test.strictSame(items.sort(), files);
    metasync.each(
      items,
      (item, cb) => {
        if (item.includes('dir')) testTestDir(path.join(current, item), test, cb);
        else {
          cb();
          return;
        }
      },
      cb
    );
  });
};
```

test/fs.js
KHYehor: I have read about this and tested, but I have 6 different dirs names and I can't just use one beforeEach+afterEach for all the, because I should change dir name srcDir + 'x'
or I can do my tests consistently

``` javascript
fsTest.beforeEach((test, cb) => {
  createTestDir(srcDir + '1', err => {
    if (err) {
      test.error(err);
      return;
    }
    cb();
  });
});
fsTest.afterEach((test, cb) => {
  rmdirp(srcDir + '1', err => {
    if (err) {
      test.error(err);
      return;
    }
    rmdirp(targetDir + '1', err => {
      if (err) {
        test.error(err);
        return;
      }
      cb();
    });
  });
});
```

SemenchenkoVitaliy: You need only one test dir for each test. `beforeEach` and `afterEach` are created specifically for this. Also, about `createTestDir`. Why are you trying to use recursion so much when it can be easily avoided? Tests don't need to have high performance, they should be simple enough so they can be inspected in the future and they need to be an example of usage. Something like this is closer to what we need.
```js
const srcFiles = [
  'test/srcDir',
  'test/srcDir/file1',
  'test/srcDir/dir1',
  'test/srcDir/dir2',
  'test/srcDir/dir1/subdir1',
];

const dstFiles = [
  'test/dstDir',
  'test/dstDir/file1',
  'test/dstDir/dir1',
  'test/dstDir/dir2',
  'test/dstDir/dir1/subdir1',
];

const copydirTest = metatests.test('Recursive dir copy');
copydirTest.endAfterSubtests();

copydirTest.beforeEach((test, cb) => {
  metasync.series(srcFiles, (file, cb) => {
      if (path.basename(file).startsWith('file')) {
        fs.appendFile(file, testData, cb);
      } else {
        fs.mkdir(file, cb);
      }
    }, test.cb(cb)
  );
});

copydirTest.afterEach((test, cb) => {
  metasync.series(common.merge(srcFiles, dstFiles).reverse(), (file, cb) => {
      if (path.basename(file).startsWith('file')) {
        fs.unlink(file, cb);
      } else {
        fs.rmdir(file, cb);
      }
    }, test.cb(cb)
  );
});

copydirTest.test('subtest 1', test => {
  ...
}

copydirTest.test('subtest 2', test => {
  ...
}
```
To be clear: this is not the only way to do this, that is just an example, but it is more readable and more efficient than the existing implementation. If you have questions about how [@metarhia/metatests](https://github.com/metarhia/metatests) are used you can check out existing tests in this or other repositories.

test/fs.js
SemenchenkoVitaliy: Same here: why recursion? Almost every recursion can be rewritten to loop and will look much more readable. As for your question: with https://github.com/metarhia/common/pull/251#discussion_r258946046 it can be reduced to this:
```js
copydirTest.test('test 1', test => {
  copyDir(srcFiles[0], dstFiles[0], err => {
    if (err) {
      test.error(err);
      test.end();
      return;
    }
    metasync.each(dstFiles, fs.access, test.cb(() => test.end()));
  });
});
```
Again, this is only one approach to do this and it is can be done differently.

test/fs.js
KHYehor: Yes, I just don't know the whole potential/usage of metatests.

test/fs.js
KHYehor: Hm, It looks definitely better. But still, If I will do 6 tests for the same ```srcDir``` ```dstDir``` it will cause error because I am trying to create/delete/copy/check the same dir 6 times.

KHYehor: How can I call next test only after previous test was finished?

SemenchenkoVitaliy: > If I will do 6 tests for the same `srcDir` `dstDir` it will cause error because I am trying to create/delete/copy/check the same dir 6 times.

`beforeEach` and `afterEach` are supposed to do exactly that: create directories before the test/clean up after the test. There is no need for a separate directory for each individual test. Tests are running sequentially by default, so there won't be race conditions and you don't have to worry about same directories accessed by several tests at the same time.

> How can I call next test only after previous test was finished?

Tests are executed by [@metarhia/metatests](https://github.com/metarhia/metatests) [`runner`](https://github.com/metarhia/metatests/blob/master/lib/runner.js). Subtests are instances of `ImperativeTest` and behave just like usual `metatests.test`.

package.json
```diff
@@ -49,7 +49,7 @@
     "eslint-plugin-prettier": "^3.0.1",
     "metaschema": "0.0.28",
     "metasync": "^0.3.31",
-    "metatests": "^0.3.0",
+    "metatests": "^0.5.0",
```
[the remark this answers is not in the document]
KHYehor: So, I shouldn't to change it? Left ^0.3.0 version?

test/fs.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+    metasync.each(dstFiles, fs.access, test.cb(() => test.end()));
+  });
+});
+
+copydirTest.test('transform', test => {
+  copyDir(
+    srcFiles[0],
+    dstFiles[0],
+    {
```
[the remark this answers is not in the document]
KHYehor: So, move all options for all tests, or only for this test?

package.json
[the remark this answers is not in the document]
SemenchenkoVitaliy: It should be `5.0.0` in order to use `test.cb` and `test.cbFail`, but you don't have to change it yourself. You have to [rebase](https://git-scm.com/docs/git-rebase) on master and resolve conflicts if there are any. To make it a little easier I suggest to squash commits before this.

test/fs.js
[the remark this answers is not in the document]
SemenchenkoVitaliy: Only for this.

KHYehor:
So, Is it what I should to do?

SemenchenkoVitaliy:
@KHYehor After squashing you need to rebase on `origin/master`.
```
git fetch
git rebase origin/master
```

KHYehor:
@SemenchenkoVitaliy I should do on master branch ```git rebase recursiveDirCopy``` to merge it, or what?

test/fs.js
```diff
@@ -0,0 +1,171 @@
...
+  'dstDir/dir1/subdir1',
+];
+
+class ToUpperCase extends stream.Transform {
+  constructor(options = {}) {
+    options.decodeStrings = false;
+    super(options);
+  }
+  _transform(chunk, encoding, callback) {
+    if (encoding !== 'utf8') {
```
SemenchenkoVitaliy: According to [stream documentation](https://nodejs.org/api/stream.html#stream_transform_transform_chunk_encoding_callback), it can be done much simpler.
```js
if (encoding === 'utf8') callback(null, chunk.toUpperCase());
else callback(new Error('Only utf8 is supported'));
```

test/fs.js
```diff
@@ -0,0 +1,171 @@
...
+});
+
+copydirTest.test('dereference', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], { dereference: true }, err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(dstFiles, fs.access, test.cb(() => test.end()));
```
SemenchenkoVitaliy: You should check if dereferenced files also exist.

test/fs.js
```diff
@@ -0,0 +1,171 @@
...
+});
+
+copydirTest.test('transform', test => {
+  common.copyDir(srcFiles[0], dstFiles[0], transOptions, err => {
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(dstFiles, fs.access, test.cb(() => test.end()));
```
SemenchenkoVitaliy: There should be check if contents of copied files were transformed.

test/fs.js
```diff
@@ -0,0 +1,171 @@
...
+    }
+    metasync.each(
+      common.merge(dstFiles.slice(0, 3), dstFiles.slice(-2)),
+      fs.access,
+      test.cb(() => test.end())
+    );
+  });
+});
+
+copydirTest.test('depth more expected', test => {
```
```suggestion
copydirTest.test('depth more than expected', test => {
```

test/fs.js
```diff
@@ -0,0 +1,171 @@
...
+    if (err) {
+      test.error(err);
+      test.end();
+      return;
+    }
+    metasync.each(dstFiles, fs.access, test.cb(() => test.end()));
+  });
+});
+
+copydirTest.test('depth less expected', test => {
```
```suggestion
copydirTest.test('depth less than expected', test => {
```

lib/fs.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const metasync = require('metasync');
+
+const error = (err, options, cb) => {
+  if (options.errorLog) {
+    options.errorLog.write(err.stack);
+    cb();
```
SemenchenkoVitaliy: IFAIU, the error should be always passed to a callback and duplicated to an `options.errorLog` if it is specified.

test/fs.js
KHYehor: To be honest I just took Alexey Orlando example stream.
Ok I will remake it.

SemenchenkoVitaliy:
> I should do on master branch `git rebase recursiveDirCopy` to merge it, or what?

First of all, you shouldn't push to `master` if tests fail. Second, you cannot push to protected branches yourself, every repo has its [main maintainer](https://github.com/metarhia/Metarhia/blob/master/doc/maintainers.md) who will do this.

test/fs.js
SemenchenkoVitaliy: You cannot just copy the code of others without permission or giving them credits for that. In case this is necessary you should check if [the licence](https://choosealicense.com/appendix/) allows that and give credit to the original author. If this is not done it can create problems with copyright. Furthermore, this is not a complex code and it shouldn't be copied.

lib/fs.js
KHYehor: Like this?
```javascript
  if (options.errorLog) {
    options.errorLog.write(err.stack);
  } 
  cb(err);
```

SemenchenkoVitaliy: Yes, though you can make an if statement one-liner.
```js
  if (options.errorLog) options.errorLog.write(err.stack);
  cb(err);
```

test/fs.js
KHYehor: hmmm, Can you suggest me something?
```fs.readFile``` than compare it with ```.toUperCase()```?

Comment (SemenchenkoVitaliy) [reply test/fs.js]:
```js
metasync.each(dstFiles,
  (file, cb) => {
    const fileName = path.basename(file);
    if (fileName.startsWith('file')) {
      fs.readFile(file, (err, data) => {
        test.strictSame(data, expectedData);
        cb(err);
      });
    } else if (fileName.startsWith('link')) {
      // fs.access or fs.readFile depending on whether this is a file or directory
    } else {
      fs.access(file, cb);
    }
  },
  test.cb(() => test.end())
);
```

test/fs.js
```diff
@@ -0,0 +1,171 @@
...
+const copydirTest = metatests.test('Recursive dir copy');
+copydirTest.endAfterSubtests();
+
+copydirTest.beforeEach((test, cb) => {
+  metasync.series(
+    srcFiles,
+    (file, cb) => {
+      if (path.basename(file).startsWith('file')) {
+        fs.appendFile(file, testData, cb);
+      } else if (file.includes('link')) {
```
```suggestion
      } else if (path.basename(file).startsWith('link')) {
```

test/fs.js
KHYehor: I should do the almost same fn in each test
So it will be better to move it in separate function?

SemenchenkoVitaliy: Tests for `transform`, `dereference`, `filter`, `depth less than expected` should have different check functions since copying is performed differently, therefore, there is no need to create a separate function for this.

test/fs.js
KHYehor: Just explain me, why do you do it this way?

SemenchenkoVitaliy: This is related to https://github.com/metarhia/common/pull/251#discussion_r260271715. For example, this should be valid.
```js
const srcFiles = [
  'linkDir1',
  'srcDir',
  'srcDir/link1', // it will be linked to `linkDir1`,
  'srcDir/link1/file1', // it also will be linked to `linkDir1`,
];
```
It will make directory `linkDir1`, create a symlink from `srcDir/link1` to it and create a file `file1`. And though it doesn't make much sense in this context as it will be much better to have something like this.
```js
const srcFiles = [
  'linkDir1',
  'linkDir1/file1',
  'srcDir',
  'srcDir/link1', // it will be linked to `linkDir1`,
];
```
Also, `path.basename` and `String.prototype.startsWith` should work faster than `String.prototype.includes` if you put `path.basename(file)` into a separate variable though v8 might optimise it. You can learn more about it [here](https://github.com/nodejs/node/blob/master/lib/path.js), [here](https://chromium.googlesource.com/v8/v8/+/refs/heads/7.0.276/src/builtins/builtins-string.cc) and [here](https://chromium.googlesource.com/v8/v8/+/refs/heads/7.0.276/src/builtins/builtins-string-gen.cc).
Another thing is code style consistency: having similar conditions will make it easier to read.

SemenchenkoVitaliy: And a few more things about links. Now if `dereference` option is true and there are recursive symlinks `copyDir` will fail.
```js
const srcFiles = [
  'srcDir/dir1',
  'srcDir/dir1/link2', // linked to `dir1`,
// or
  'srcDir/dir2',
  'srcDir/dir3',
  'srcDir/dir2/link3', // linked to `dir3`,
  'srcDir/dir3/link4', // linked to `dir2`,
];
```
This is something that we don't need right now, but if you feel about it you can try to do this.

SemenchenkoVitaliy:
ping @belochub @nechaido @lundibundi.

KHYehor:
@SemenchenkoVitaliy So everything is okey?

SemenchenkoVitaliy:
@KHYehor Please, wait until other members of [@metarhia](https://github.com/metarhia) take a look at this PR. Thank you for your patience. Also, we can't land this before https://github.com/metarhia/common/pull/255.

KHYehor:
@SemenchenkoVitaliy if it wouldn't be so hard for you, could you tell me my mistakes and suggest me something about writing code?

SemenchenkoVitaliy:
@KHYehor Less is more and simpler is better. Sometimes it is better to write code that may not perform better but it will be more readable and can be changed easier. There are many [@metarhia](https://github.com/metarhia) libraries, so don't hesitate to check them out and ask a question if something is unclear. As for the code style, you can check out [mdn](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style), [google](https://google.github.io/styleguide/javascriptguide.xml#for-in_loop) style guides or take a look at the code in this or other repositories of [@metarhia](https://github.com/metarhia). Hope this was helpful :)

belochub:
@KHYehor rebase on the `master` branch, please.

SemenchenkoVitaliy:
@KHYehor Will you work on this further?

KHYehor:
> @KHYehor Will you work on this further?

More likely no than yes, because this function most likely appear in `lib 'fs'` as it was with recursive dir deletion function.

PR #252, opened 2019-02-13
Implement recursive rmdir

JuliaGerasymenko:


common.js
```diff
@@ -11,6 +11,7 @@ const submodules = [
   'events', // Events and emitter
   'flags', // Flags data type
   'fp', // Functional programming
+  'fs', // Recursive mkdir
```
SemenchenkoVitaliy:
```suggestion
  'fs', // file system utilities
```

lib/fs.js
```diff
@@ -0,0 +1,61 @@
...
+          else fs.rmdir(dir, cb);
+        });
+      });
+    } else {
+      fs.unlink(dir, cb);
+    }
+  });
+};
+
+const mkdir = (dir, mode, cb) => {
```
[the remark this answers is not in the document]
nechaido: AFAIU rmdir and mkdirp use the other function for tests, because of this it is hard to make them truly independent.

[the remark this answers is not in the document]
SemenchenkoVitaliy: Yes, but function tests shouldn't depend on other function.  For  example, test for `rmdirp` can look something like this:
```js
metatests.test('rmdirp test', test =>
  fs.mkdir('./testDir', err =>
    fs.mkdir('./testDir/subdir1', err =>
      fs.mkdir('./testDir/subdir1/subdir2',
        rmdirp('./testDir', err =>
          fs.access('./testDir', err => {
            test.isError(err);
            test.end();
          })
        )
      )
    )
  )
);
```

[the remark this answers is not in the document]
nechaido: @SemenchenkoVitaliy It's not that simple: if a directory exists `mkdir` will result in an error. You should also check an error in case it is a real fs access error, which will transform your test into:
```js
metatests.test('rmdirp test', test =>
  fs.mkdir('./testDir', err => {
    if (err && err.code !== 'EEXISTS') {
      test.bailout(err);
    }
    fs.mkdir('./testDir/subdir1', err => {
      if (err && err.code !== 'EEXISTS') {
        test.bailout(err);
      }
      fs.mkdir('./testDir/subdir1/subdir2', err => {
        if (err && err.code !== 'EEXISTS') {
          test.bailout(err);
        }
        rmdirp('./testDir', err => {
          if (err && err.code !== 'EEXISTS') {
            test.bailout(err);
          }
          fs.access('./testDir', err => {
            test.isError(err);
            test.end();
          });
        });
      });
    });
  })
);

```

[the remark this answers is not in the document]
SemenchenkoVitaliy: Of course, there should be errors checks. I just proposed a concept code. As you can see, it doesn't even have error handling. Your example is something that should actually be used.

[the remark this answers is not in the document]
SemenchenkoVitaliy: And of course, we should make cleanup after test if it fails.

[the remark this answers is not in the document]
nechaido: > Your example is something that should actually be used.

I don't feel like it because there is a lot of boilerplate code it that case, and if we try to make some helper function we'll end up implementing both `mkdirp` and `rmdir`.

> And in case if the test fails, cleanup should be performed.

It's not always possible. The only way for preparation to fail is to have some issues with the filesystem, in that case cleanup is useless, because it will probably fail too.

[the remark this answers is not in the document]
SemenchenkoVitaliy: I think that if we have only 2 levels of nesting i.e. `./testDir` and `./testDir/subdir1` it won't look too bad. What I want to say is that tests for `mkdipr` shouldn't use `rmdirp` and vise versa. Yeah, sure, tests will look awful but I don't think we can do something about this. Moreover, `mkdirp` and `rmdipr` have their own PRs. Of course, we can create new PR, which will implement both `mkdirp` and `rmdirp` but IMO we should land those PRs for each function separately with their own dedicated tests and then fix tests or add more if it is needed.

SemenchenkoVitaliy:
@JuliaGerasymenko Commit should be renamed to `Implement recursive rmdir`. Also, please, don't rebase commits each time you make changes. Use `git commit --fixup` for this purpose.

test/fs.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const metatests = require('metatests');
+const fs = require('fs');
+const { rmdirp } = require('..');
+
```
SemenchenkoVitaliy: The test directory path should be moved in the constant at the beginning and I think it should be `test/testDir`, not just `testDir`. Also, we might support Windows platform in the future, so it will be more convenient to use `path.sep` or `path.join` instead of hardcoding.
```suggestion
const RMDIRP_TEST_DIR = `test${path.sep}testDir`;
```

test/fs.js
```diff
@@ -0,0 +1,27 @@
...
+metatests.test('rmdirp test', test =>
+  fs.mkdir('./testDir', err => {
```
```suggestion
  fs.mkdir(RMDIRP_TEST_DIR, err => {
```

test/fs.js
```diff
@@ -0,0 +1,27 @@
...
+    if (err && err.code !== 'EEXISTS') {
+      test.bailout(err);
+    }
+    fs.mkdir('./testDir/subdir1', err => {
```
```suggestion
    fs.mkdir(path.join(RMDIRP_TEST_DIR, 'subdir1'), err => {
```

test/fs.js
```diff
@@ -0,0 +1,27 @@
...
+      if (err && err.code !== 'EEXISTS') {
+        test.bailout(err);
+      }
+      rmdirp('./testDir', err => {
```
```suggestion
      rmdirp(RMDIRP_TEST_DIR, err => {
```

test/fs.js
```diff
@@ -0,0 +1,27 @@
...
+        if (err && err.code !== 'EEXISTS') {
+          test.bailout(err);
+        }
+        fs.access('./testDir', err => {
```
```suggestion
        fs.access(RMDIRP_TEST_DIR, err => {
```

lib/fs.js
```diff
@@ -0,0 +1,34 @@
...
+      else cb(err);
+      return;
+    }
+    if (stats.isDirectory()) {
+      fs.readdir(dir, (err, files) => {
+        if (err) {
+          cb(err);
+          return;
+        }
+        files = files.map(file => path.join(dir, file));
```
tshemsedinov:
```suggestion
        const paths = files.map(file => path.join(dir, file));
```
I think it's better not to change args

lib/fs.js
```diff
@@ -0,0 +1,34 @@
...
+        metasync.each(files, rmdirp, err => {
```
tshemsedinov:
```suggestion
        metasync.each(paths, rmdirp, err => {
```

lib/fs.js
```diff
@@ -0,0 +1,34 @@
...
+          if (err) cb(err);
+          else fs.rmdir(dir, cb);
+        });
+      });
+    } else {
+      fs.unlink(dir, cb);
```
tshemsedinov: `rmdirp(fileNotDir, ....)` should return `ENOTDIR` as in `fs.rmdir`

lib/fs.js
```diff
@@ -17,6 +18,22 @@ const mkdir = (dir, mode, cb) => {
   });
 };
 
+const recursivlyListDirs = dir => {
+  const list = [dir];
+  let nextDir = dir;
+  const root = dir.startsWith('/') ? '/' : '.';
```
belochub: To make this OS agnostic:
```suggestion
  const root = path.parse(dir).root || '.';
```

test/fs.js
```diff
@@ -34,3 +35,25 @@ testMkdirp.test('create 2 directories without mode', test => {
...
+  fs.mkdir(RMDIRP_TEST_DIR, err => {
+    if (err && err.code !== 'EEXISTS') {
+      test.bailout(err);
+    }
+    fs.mkdir(path.join(RMDIRP_TEST_DIR, 'subdir1'), err => {
+      if (err && err.code !== 'EEXISTS') {
+        test.bailout(err);
+      }
+      rmdirp(path.join(RMDIRP_TEST_DIR, 'subdir1'), err => {
+        if (err && err.code !== 'EEXISTS') {
```
belochub: I think `EEXISTS` should not be ignored here:  
```suggestion
        if (err) {
```

lib/fs.js
```diff
@@ -17,6 +18,22 @@ const mkdir = (dir, mode, cb) => {
...
+  while ((nextDir = path.dirname(nextDir)) !== root) list.push(nextDir);
+  return list;
+};
+
+const rmdirp = (dir, cb) => {
+  metasync.series(
+    recursivlyListDirs(dir),
+    (dirName, cb) => fs.rmdir(dirName, cb),
```
belochub: I don't think we should return the value obtained from calling the `fs.rmdir()` here:
```suggestion
    (dirName, cb) => {
      fs.rmdir(dirName, cb);
    },
```

lib/fs.js
```diff
@@ -17,6 +18,22 @@ const mkdir = (dir, mode, cb) => {
...
+    err => cb(err)
```
belochub: Same here:
```suggestion
    err => {
      cb(err);
    }
```

lib/fs.js
SemenchenkoVitaliy: What about just `fs.rmdir`?
```js
    fs.rmdir,
```

lib/fs.js
```diff
@@ -17,6 +18,20 @@ const mkdir = (dir, mode, cb) => {
   });
 };
 
+const recursivlyListDirs = dir => {
```
SemenchenkoVitaliy: It has incorrect behaviour. It should list directories which are nested in `dir`, not vice versa.

test/fs.js
```diff
@@ -34,3 +35,25 @@ testMkdirp.test('create 2 directories without mode', test => {
...
+metatests.test('rmdirp test', test =>
+  fs.mkdir(RMDIRP_TEST_DIR, err => {
+    if (err && err.code !== 'EEXISTS') {
+      test.bailout(err);
+    }
+    fs.mkdir(path.join(RMDIRP_TEST_DIR, 'subdir1'), err => {
+      if (err && err.code !== 'EEXISTS') {
+        test.bailout(err);
+      }
+      rmdirp(path.join(RMDIRP_TEST_DIR, 'subdir1'), err => {
```
SemenchenkoVitaliy: It should be
```suggestion
      rmdirp(RMDIRP_TEST_DIR, err => {
```

Issue #253, opened 2019-02-13
Version check for polyfills


SemenchenkoVitaliy:
Sometimes we need to implement polyfills for older Node.js versions and for this we have to check version manually like:
```js
const vArr = process.versions.node.split('.').map(num => parseInt(num));
if (vArr[0] > 8 || (vArr[0] === 8 && vArr[1] >= 2)) {
  ...
}
```
I propose to implement function that will check if process' version is older than required.
```js
if (common.nodeVersionCompatible('8.2')) {
  ...
}
```
or
```js
if (common.v8VersionCompatible('8.2')) {
  ...
}
```
WDYT?
/cc @lundibundi @nechaido @belochub @tshemsedinov

nechaido:
AFAIK we need to check that the current version is equal or higher than the specified version, not compatible.
Example: we only use mkdip polyfill for <10.12.0.
 
| Version | Polyfill | v10.12.0 compatible |
|---------|---------|------------|
| 8.0.0   | :heavy_check_mark:  | :negative_squared_cross_mark:  |
| 10.0.0  | :heavy_check_mark:  | :negative_squared_cross_mark:  |
| 10.12.0 | :negative_squared_cross_mark:  | :heavy_check_mark:  |
| 11.0.0  | :negative_squared_cross_mark:  | :negative_squared_cross_mark:  |

SemenchenkoVitaliy:
@nechaido Agree, function name should be changed. What about `verifyVersion` or `checkVersion`? Can as well make `compareVersions`.

PR #254, opened 2019-02-14
Implement version check and compare functions

SemenchenkoVitaliy:
Closes: https://github.com/metarhia/common/issues/253

/cc @lundibundi @nechaido @belochub


belochub:
Can't we just use [the `semver` package](http://npmjs.com/semver) for this?

SemenchenkoVitaliy:
@belochub This would bring new external dependencies.

belochub:
@SemenchenkoVitaliy and what's the problem with that?

SemenchenkoVitaliy:
@belochub I thought that we will remove most external dependencies from Metarhia stack and replace them with our own in the future. If this is not the case then I think that this PR can be closed.

tshemsedinov:
I think it will be better to remove `semver` dependency and any other dependencies in future. We are going to create as independent infrastructure as possible. We may use `devDependencies` without limitations but if we can create replacement of `dependencies` quickly it is preferable.

belochub:
Closing for now since there is no need to add this kind of public API to this package at the moment.

PR #257, opened 2019-03-07
Add implementation of pushSame for array

belochub:
I've decided to create this function after benchmarking it against the alternatives:
https://jsbench.me/pzjsyqrz9h/1

### DIFF
--- a/lib/array.js
+++ b/lib/array.js
@@ -64,10 +64,26 @@ const sequence = (seq, max) => {
 // Returns: <any>, element
 const last = arr => arr[arr.length - 1];
 
+// Push single value multiple times
+//   arr - <Array>
+//   n - <number>
+//   value - <any>
+// Returns: <number>, new value of arr.length
+const pushSame = (arr, n, value) => {
+  if (n <= 0) return arr.length;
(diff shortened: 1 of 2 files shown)


lib/array.js
```diff
@@ -64,10 +64,26 @@ const sequence = (seq, max) => {
 // Returns: <any>, element
 const last = arr => arr[arr.length - 1];
 
+// Push single value multiple times
+//   arr - <Array>
+//   n - <number>
+//   value - <any>
+// Returns: <number>, new value of arr.length
+const pushSame = (arr, n, value) => {
```
nechaido: `pushMultiple`?

belochub: @nechaido the function pushes the same value 0 or more times, thus I think naming it `pushMultiple` would be ambiguous.

nechaido: @belochub I do not like `pushSame`, but I do not have any better option, so you can proceed and land it.

belochub:
Landed in https://github.com/metarhia/common/commit/54731888dfcef45946cba5a671d18ea9a4998974.

PR #258, opened 2019-03-12
Expose AuthenticationStrength.compliance instead of strength

lundibundi:


test/auth.js
```diff
@@ -232,26 +232,26 @@ metatests.test('test AuthenticationStrength / strength', test => {
   ];
 
   let result = common.checkPassword(password, required, optional);
-  test.strictSame(result.strength, 'Not valid');
+  test.strictSame(result.compliance, 0.25);
```
SemenchenkoVitaliy: I think there should be also
```js
test.assertNot(result.valid);
```

SemenchenkoVitaliy: And on the other tests as well.

PR #259, opened 2019-03-21
Fix `duplicateWithReferences`

mille-nium:
Previous version won't work with objects, that have no prototype, as they don't have `hasOwnProperty` method


belochub:
Landed in https://github.com/metarhia/common/commit/d6edf01ce802518e5db257b20de943ae4fef15cd.

PR #261, opened 2019-03-26
Add password auth tests

bugagashenkj:


lib/auth.js
```diff
@@ -76,22 +103,25 @@ const passwordTests = {
...
-      // https://www.owasp.org/index.php/Password_special_characters
-      return stringIncludesChars(password, SPECIAL_CHARS_RANGE, options.number);
-    },
+    test: (password, options) =>
+      stringIncludesChars(password, SPECIAL_CHARS_RANGE, options.number),
     hint: options => ({ name: 'MIN_SPECIAL_CHARS', number: options.number }),
     options: { number: 1 },
   },
+  FOUND_TOPOLOGY: {
+    test: password => isTopology(password),
```
lundibundi: I think we should allow to pass custom list of topologies (but still have our default `passwordTopologies`).

lundibundi: Why not make `passwordTopologies` a Set? (as well as `popularPasswords` below)

PR #262, opened 2019-03-27
Fix Iterator#includes() for non-number values

belochub:


belochub:
Landed in https://github.com/metarhia/common/commit/1169f56a873da077d29641d89df4e91d310a56da.

PR #263, opened 2019-03-27
Version 1.4.0

belochub:
* Implement recursive rmdir
* Fix Iterator#includes() for non-number values
* Add password auth tests


belochub:
Landed in https://github.com/metarhia/common/commit/2a8e362a299d79793306472b5872fcff12b38de0.

PR #264, opened 2019-03-27
Fix dependencies

belochub:
Fix cyclic dependency on metasync and add missing metasync dependency to
package.json.


belochub:
Landed in https://github.com/metarhia/common/commit/9b50d8e39961b6b8355d8490f08278919f9bd131.

PR #267, opened 2019-03-28
Version 1.4.2

belochub:
* Remove metasync dependency to avoid issues with circular `require()`s


belochub:
Landed in https://github.com/metarhia/common/commit/76ea995b6f183c4515485cdf7b36549d48e9e99a.

PR #283, opened 2019-04-24
Fix documentation

SemenchenkoVitaliy:


README.md
```diff
@@ -398,8 +398,6 @@ Merge multiple objects with merger
 
 #### Enum.from(...args)
 
-#### Enum.NaE()
```
belochub: I think this should be fixed to show that it is not a method but a property, instead of deleting it.

belochub:
Landed in https://github.com/metarhia/common/commit/e5d0b28d2443e75d325b27b3c587f1e22db73cd2.

PR #286, opened 2019-04-25
Replace Math.pow() with exponentiation operator

belochub:


belochub:
Landed in https://github.com/metarhia/common/commit/f7e72c8f0f62834898d8906c81566cb116e7c80b.

PR #287, opened 2019-04-25
Use String#padStart() instead of manual padding

belochub:


belochub:
Landed in https://github.com/metarhia/common/commit/2ad440927a1480673c6594bec15b553c9bc2bd8f.

PR #290, opened 2019-05-16
Impl recursive rm

SemenchenkoVitaliy:
Info: https://github.com/metarhia/filestorage/pull/12#discussion_r283781232

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -1,13 +1,15 @@
 'use strict';
 
 const fs = require('fs');
-const path = require('path');
+const pathModule = require('path');
+const util = require('util');
+const { iter } = require('./iterator');
 const MKDIRP_DEFAULT_MODE = 0o777;
 
 const mkdir = (dir, mode, cb) => {
   fs.access(dir, fs.constants.F_OK, err => {
     if (err && err.code === 'ENOENT') {
-      mkdir(path.dirname(dir), mode, err => {
+      mkdir(pathModule.dirname(dir), mode, err => {
         if (err) cb(err);
         else fs.mkdir(dir, mode, cb);
       });
@@ -20,8 +22,8 @@ const mkdir = (dir, mode, cb) => {
 const recursivelyListDirs = dir => {
   const list = [dir];
   let nextDir = dir;
-  const root = path.parse(dir).root || '.';
-  while ((nextDir = path.dirname(nextDir)) !== root) list.push(nextDir);
+  const root = pathModule.parse(dir).root || '.';
+  while ((nextDir = pathModule.dirname(nextDir)) !== root) list.push(nextDir);
   return list;
 };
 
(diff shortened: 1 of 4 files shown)


lib/fs.js
```diff
@@ -62,7 +64,72 @@ if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
...
+  fsPromises = iter(list).collectWith(
+    {},
+    (obj, name) => (obj[name] = util.promisify(fs[name]))
+  );
+}
+
+// Recursively remove directory
+//   file <string> file or directory to be removed
+const rmRecursivePromise = async file => {
+  const stats = await fsPromises.stat(file);
```
belochub: Is there actually a need to `fs.stat()` the file/directory?
Can't we just use `fs.readdir()`, then check if the error is `ENOTDIR` and perform `fs.unlink()` in that case?

lib/fs.js
```diff
@@ -62,7 +64,72 @@ if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
...
+      cb(err);
+      return;
+    }
+    if (stats.isDirectory()) {
+      fs.readdir(file, (err, files) => {
+        if (err) {
+          cb(err);
+          return;
+        }
+        const rm = files.reduce(
```
belochub: Can you also rewrite it to run multiple parallel asynchronous operations instead of sequential execution?

lib/fs.js
```diff
@@ -62,7 +64,72 @@ if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
   };
 }
 
+// Recursively remove directory
+//   file <string> file or directory to be removed
```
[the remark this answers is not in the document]
SemenchenkoVitaliy: Yes, but in this case, the argument in the comment will divorce with the argument in the code, and I'd prefer to leave them the same.

lib/fs.js
SemenchenkoVitaliy: IMO, it will look less readable. Compare
```js
  const stats = await fsPromises.stat(file);
  if (stats.isDirectory()) {
    const files = await fsPromises.readdir(file);
    await Promise.all(files.map(f => rmRecursivePromise(path.join(file, f))));
    return fsPromises.rmdir(file);
  } else {
    return fsPromises.unlink(file);
  }
```
and

```js
  let files;
  try {
    files = await fsPromises.readdir(file);
  } catch (err) {
    if (err.code === 'ENOTDIR') {
      return fsPromises.unlink(file);
    } else {
      throw err;
    }
  }
  await Promise.all(files.map(f => rmRecursivePromise(path.join(file, f))));
  return fsPromises.rmdir(file);
```

belochub: You can write it the following way:
```js
const rmRecursivePromise = async file =>
  fsPromises.readdir(file).then(async files => {
    await Promise.all(files.map(f => rmRecursivePromise(path.join(file, f))));
    return fsPromises.rmdir(file);
  }, err => {
    if (err.code === 'ENOTDIR') return fsPromises.unlink(file);
    throw err;
  }); 
```

This is both shorter and, in my opinion, easier to understand, than using try/catch.

lib/fs.js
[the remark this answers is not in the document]
belochub: No, I meant changing the argument name in the comment as well as in code, not just in the comment.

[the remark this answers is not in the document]
SemenchenkoVitaliy: Well, in this case, we will need to rename the existing `path` variable which is Node.js [`path`](https://nodejs.org/api/path.html) library to something like `nodePath`. If you can suggest a better name, I'll change it, but I'm strongly against having the same library not only under a different name than a library name but also different across different files.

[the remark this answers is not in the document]
belochub: Just call it `pathModule` like it is called in the Node.js source code:
https://github.com/nodejs/node/blob/6f55c5327ebb0ff169a73c698168a7fb4cd604bc/lib/fs.js#L44

The reasoning to do so is to keep the signature the same as [`fs.unlink()`](https://nodejs.org/api/fs.html#fs_fs_unlink_path_callback) and [`fs.rmdir()`](https://nodejs.org/api/fs.html#fs_fs_unlink_path_callback).

lib/fs.js
SemenchenkoVitaliy: BTW, shouldn't we also support Windows, since this code will be used in [@metarhia/filestorage](https://github.com/metarhia/filestorage) and on Windows error code will be `ENOENT`?

belochub: @SemenchenkoVitaliy, yes, sure.

lib/fs.js
```diff
@@ -62,7 +64,79 @@ if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
...
+const rmRecursivePromise = async path =>
+  fsPromises.readdir(path).then(
+    async files => {
+      await Promise.all(
+        files.map(f => rmRecursivePromise(pathModule.join(path, f)))
+      );
+      return fsPromises.rmdir(path);
+    },
+    err => {
+      if (err.code === 'ENOTDIR' || err.code === 'ENOENT') {
```
belochub: Maybe add a check to avoid skipping `ENOENT` on *nix systems?
```suggestion
      if (err.code === 'ENOTDIR' || process.platform === 'win32' && err.code === 'ENOENT') {
```

lib/fs.js
```diff
@@ -62,7 +64,73 @@ if (version[0] < 10 || (version[0] === 10 && version[1] <= 11)) {
...
+  fsPromises = iter(['readdir', 'rmdir', 'unlink']).collectWith(
+    {},
+    (obj, name) => {
+      obj[name] = util.promisify(fs[name]);
+    }
+  );
+}
+
+// Recursively remove directory
+//   path <string> path to a file or directory to be removed
```
belochub: I forgot, there should also probably be the `Returns:` comment specified:
```js
// Returns: <Promise>
```

belochub:
Landed in https://github.com/metarhia/common/commit/0b2ec8d761cade36ad9c19659778a2fcf5ce6978.

PR #294, opened 2019-06-01
Implement Iterator.filterMap enabling more efficient filter+map

lundibundi:
It will call mapper on every element just like .map but will also
filter out values returned from mapper that match
filterValue (undefined by default).

- [x] tests are added/updated


belochub:
Landed in https://github.com/metarhia/common/commit/83ffb0442f7931f46ac56447904f99b06a322a37.

PR #301, opened 2019-07-01
Add, refactor, and fix tests

belochub:
* Improve coverage for most of the modules, viz.: auth.js, cache.js, callbacks.js, data.js, fp.js, id.js, math.js, network.js, oop.js, stream.js, strings.js, time.js, and utilities.js
* Fix invalid argument order in assertions.
* Remove outdated API usage.
* Move all of the tests scattered across the multiple files separated by function names into the files named after the modules that are being tested.

This PR improves the overall code coverage up to 99% with the only not 100% covered module being `fs.js`.

- [x] tests are added/updated


test/time.js
```diff
@@ -3,6 +3,22 @@
...
+    return DefaultDate.apply(DefaultDate, args);
+  }
+  if (args.length !== 0) {
+    return new DefaultDate(...args);
+  }
+  return new DefaultDate(dateNowValue);
+};
+Date.now = function now() {
+  return dateNowValue;
+};
```
lundibundi: I assume this is needed to check `if (!date) date = new Date();` branch in those functions but can't we just generate the correct date on the fly?
i.e. 
```js
`${date.getUTCFullYear()}-${(date.getUTCMonth() + 1).toString().padStart(2, '0')}-${date.getUTCDate().toString().padStart(2, '0')}`
```
I know that this is code duplication of what's already present in `nowDate` and `nowDateTime` but I'd prefer that to global state modification.

belochub: That will possibly be a flaky test, and apart from that, it's the same code used to generate the tested value, which I don't think is a valid way to test it.

test/time.js
```diff
@@ -3,6 +3,22 @@
 const metatests = require('metatests');
 const common = require('..');
 
+const DefaultDate = Date;
+const dateNowValue = 1561975200000; // 2019-07-01T10:00:00.000Z
+// eslint-disable-next-line no-global-assign
+Date = function(...args) {
```
nechaido: I don't like the idea of mutating global state, currently `metatests` doesn't isolate tests, this will lead to unexpected behavior in every other test.
Currently I see 2 ways to fix it:
* Add sandboxing to metatests.
* Move this test to `utils` or `fixtures` and spawn it as a separate process.

/cc @lundibundi

belochub: @nechaido as a temporary workaround, I've moved the test to a different directory where each test file is then run sequentially in a separate process.

belochub:
Landed in https://github.com/metarhia/common/commit/258f900a57d7dff7a03e3dd2abcd98e03a6bc865.

PR #303, opened 2019-07-03
Add Iterator#chunks to allow splitting iterator values into Arrays

lundibundi:
- [x] tests are added/updated


lib/iterator.js
```diff
@@ -199,6 +199,19 @@ class Iterator {
     return result + suffix;
   }
 
+  // Split iterator into chunks of size `from` or every time function
+  // `from` returns falsy value.
+  //   from - <Function> | <number> split function or chunk size
+  //     val <any> current iterator value
+  chunks(from) {
```
nechaido: IMHO, this should use a verb for a name.

lundibundi: So how about `chunk` or `chunkBy`?

nechaido: Either one is ok with me.

lib/iterator.js
```diff
@@ -467,6 +480,26 @@ class SkipWhileIterator extends Iterator {
...
+  constructor(base, fn) {
+    super(base);
+    this.fn = fn;
+    this.value = [];
+  }
+
+  next() {
+    let next = this.base.next();
+    if (next.done) return next;
+    while (!next.done && (!this.fn(next.value) || this.value.length === 0)) {
```
nechaido: Extract `this.value.length === 0` to `firstIteration`?

test/iterator.js
```diff
@@ -610,6 +610,20 @@ metatests.testSync("Iterator.toObject with '0', '1' properties", test => {
   test.strictSame(actual, { a: 0, b: 1, c: 2 });
 });
 
+metatests.testSync('Iterator.chunks', test => {
```
nechaido: Please add tests:
* empty array
* condition that will result in one huge chunk
* `[1, 1, 3, 3, 3, 5, 5].chunks(x => x === 3)`

lib/iterator.js
lundibundi: I can't do that because it will lead to this cycle always ignoring `this.fn` calls on the first `next` iteration and therefore always consuming whole iterator on the first iteration.

test/iterator.js
```diff
@@ -643,6 +643,41 @@ metatests.testSync('Iterator.chainApply on string', test => {
...
+metatests.testSync('Iterator.chunk with empty iterator', test => {
+  const actual = iter([])
+    .chunk(3)
+    .toArray();
+  test.strictSame(actual, []);
+});
+
+metatests.testSync('Iterator.chunk with single chunk', test => {
+  const actual = iter([1, 2, 3, 1, 4, 5, 1, 6, 7])
+    .chunk(() => false)
```
SemenchenkoVitaliy: Can you also add an example with `.chunk(() => true)`?

lundibundi: Thanks, this caught a bug where the last chunk could have been ignored.

PR #305, opened 2019-07-05
Fix flaky MemoryWritable test

belochub:
It was failing on Windows due to filesystem read stream returning 61 bytes from the file that is considered 60 byte long on Unix systems.

- [x] tests are added/updated

### DIFF
--- a/test/stream.js
+++ b/test/stream.js
@@ -40,18 +40,12 @@ metatests.test(
 );
 
 metatests.test('MemoryWritable handles custom sizeLimit', async test => {
-  for (const path of fixturesPaths) {
-    const originalReadable = fs.createReadStream(path);
-    const memoryStream = new common.MemoryWritable(40);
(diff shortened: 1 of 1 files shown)


test/stream.js
```diff
@@ -40,18 +40,13 @@ metatests.test(
...
-      );
-    }
-  }
+  const totalSize = 60;
+  const limitSize = 40;
+  const memoryStream = new common.MemoryWritable(limitSize);
+  const result = memoryStream.getData();
+  memoryStream.write(Buffer.alloc(totalSize));
+  test.rejects(
+    result,
```
lundibundi: Nit: I'd prefer to request promise in a function to `rejects`:
 ```suggestion
    () => {
      const result = memoryStream.getData();
      memoryStream.write(Buffer.alloc(totalSize));
      return result;
    },
```

PR #311, opened 2019-07-11
Make CryptoRandomPrefetcher implement Iterator protocol

lundibundi:
- [x] tests are added/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to
 - `MemoryWritable` internal buffer size is now limited to 8 MB by default.
 - Signature of `callerFilepath()` to allow passing `RegExp` as depth to be used
   for filtering of stack frames.
+- Return value of `cryptoPrefetcher()` now implements the Iterable interface.
 
 ### Fixed
 
--- a/lib/math.js
+++ b/lib/math.js
@@ -23,6 +23,15 @@ class CryptoRandomPrefetcher {
     this.pos = end;
     return buf;
   }
+
+  [Symbol.iterator]() {
+    return {
+      [Symbol.iterator]() {
+        return this;
+      },
+      next: () => ({ value: this.next(), done: false }),
+    };
+  }
 }
 
 // Create prefetcher to use when crypto.randomBytes is required to generate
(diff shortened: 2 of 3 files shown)


belochub:
Why do we need this?

lundibundi:
This will allow nice interactions with our `Iterator` API 
i.e. 
```
const rand = iter(cryptoPrefetcher(40, 8));
rand.take(10).forEach(/* do smth */);
...
rand.take(5).forEach(/* do smth more*/);
```
instead of manuaaly calling `next()` each time.

And doesn't really add that much complexity/maintenance burden to us as everything is basically already there but an Iterator method itself.

test/math.js
```diff
@@ -27,3 +27,12 @@ metatests.test('cryptoPrefetcher with invalid arguments', test => {
   );
   test.end();
 });
+
+metatests.testSync('cryptoPrefetcher for of', test => {
+  const valueSize = 8;
+  const cryptoPrefetcher = common.cryptoPrefetcher(valueSize * 5, 8);
```
nechaido:
```suggestion
  const cryptoPrefetcher = common.cryptoPrefetcher(valueSize * 5, valueSize);
```

lib/math.js
```diff
@@ -23,6 +23,14 @@ class CryptoRandomPrefetcher {
     this.pos = end;
     return buf;
   }
+
+  [Symbol.iterator]() {
+    return {
```
belochub: The returned object should also include `Symbol.iterator` method.

lundibundi: Done.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to
 - `MemoryWritable` internal buffer size is now limited to 8 MB by default.
 - Signature of `callerFilepath()` to allow passing `RegExp` as depth to be used
   for filtering of stack frames.
+- `cryptoPrefetcher()` to implement Iterator protocol.
```
belochub: This is technically incorrect and misleading since now it only implements [the *Iterable* interface](https://tc39.es/ecma262/#sec-iterable-interface), not the Iterator protocol.

Besides that, I think that it's better to rewrite it as follows:
```suggestion
- Return value of `cryptoPrefetcher()` now implements the Iterable interface.
```

belochub:
Landed in https://github.com/metarhia/common/commit/d2547b9a9de7a577f8d6bff57525ce55e0156691.

PR #312, opened 2019-07-12
Add captureMaxStack utility to get maximum stack trace

lundibundi:
- [x] tests are added/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to
 - `common.mkdirpPromise()` function.
 - `Iterator#apply()` and `Iterator#chainApply()` to improve iterator
   interaction with chained calls.
+- Add `captureMaxStack()` utility to get maximum available stack trace.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to
 - `common.mkdirpPromise()` function.
 - `Iterator#apply()` and `Iterator#chainApply()` to improve iterator
   interaction with chained calls.
-- Add `captureMaxStack()` utility to get maximum available stack trace.
+- `captureMaxStack()` utility to get maximum available stack trace.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


test/utilities.js
```diff
@@ -22,3 +22,9 @@ function child(test, depth) {
   );
   test.strictSame(common.callerFilename(depth), 'utilities.js');
 }
+
+metatests.testSync('Check captureMaxStack', test => {
+  const stack = common.captureMaxStack();
+  test.assert(stack.length > 0);
+  test.type(stack, 'string');
```
belochub: Can you add some kind of regular expression check here to validate that it is a stack trace, and not some kind of random non-empty string?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to
 - `common.mkdirpPromise()` function.
 - `Iterator#apply()` and `Iterator#chainApply()` to improve iterator
   interaction with chained calls.
+- Add `captureMaxStack()` utility to get maximum available stack trace.
```
belochub:
```suggestion
- `captureMaxStack()` utility to get maximum available stack trace.
```

test/utilities.js
lundibundi: Done.

lib/utilities.js
```diff
@@ -14,6 +14,14 @@ const safe = fn => (...args) => {
   }
 };
 
+const captureMaxStack = () => {
+  const oldLimit = Error.stackTraceLimit;
+  Error.stackTraceLimit = Infinity;
+  const stack = new Error().stack;
```
SemenchenkoVitaliy: Maybe we should exclude `captureMaxStack` line from actual `stack`?

lundibundi: I don't this should deal with that, it would require processing the stack and also may be confusing to the user.
IMHO, this should be done as a separate function if needed.

belochub:
Landed in https://github.com/metarhia/common/commit/6e54d171981c1934194f256e5a7606df93ee74eb.

PR #315, opened 2019-10-11
Replace for-loop with recursion

mapogolions:
I think it's pretty weird to use construction [like this](https://github.com/metarhia/common/blob/8e755c3c793c63dc377a7607b4963c46f379f015/lib/data.js#L93) and [all of those mutable local variables](https://github.com/metarhia/common/blob/8e755c3c793c63dc377a7607b4963c46f379f015/lib/data.js#L90) to keep track a current state. I propose to use recursion. The last seems more natural. I know that JS has some problems with tail rec optimization, but i  I cannot imagine the case when a path will contain many dot-separated properties

### DIFF
--- a/lib/data.js
+++ b/lib/data.js
@@ -85,26 +85,18 @@ const getByPath = (data, dataPath) => {
 //   dataPath - <string>, dot-separated path
 //   value - <any>, new value
 const setByPath = (data, dataPath, value) => {
-  const path = dataPath.split('.');
-  const len = path.length;
-  let obj = data;
-  let i = 0;
-  let next, prop;
-  for (;;) {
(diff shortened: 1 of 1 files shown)


belochub:
Hi, @mapogolions and thanks for the contribution.
Unfortunately, I don't think that these particular functions require such substantial changes at the moment.

IIRC this code was written in the current way to run the fastest, possibly sacrificing readability, because it was used a lot in hot paths. AFAIK recursion runs notably slower than loops at the moment, so the implementation you propose will most likely hurt the performance.

I agree with the fact that these functions can be written in a more clear way and I don't mind refactoring these functions without making the switch to recursion, so feel free to open another PR with the changes.

PR #319, opened 2019-12-23
Change sortComparePriority contract

Bob-chemist:
- [x] tests are updated


Bob-chemist:
- [x] documentation is updated

CHANGELOG.md
```diff
@@ -22,8 +22,7 @@ and this project adheres to
 - `cryptoPrefetcher()` to throw when `bufSize` is not a multiple of
   `valueSize`.
 - `MemoryWritable` internal buffer size is now limited to 8 MB by default.
-- Signature of `callerFilepath()` to allow passing `RegExp` as depth to be used
-  for filtering of stack frames.
+- `sortComparePriority` function contract changed.
```
belochub: Elaborate on the change, it isn't clear in what way the contract was changed from reading that.

test/sort.js
```diff
@@ -21,19 +21,24 @@ const CONFIG_FILES_PRIORITY = [
...
     'common.sortComparePriority': [
-      [CONFIG_FILES_PRIORITY, 'files.js', 'sandbox.js', 1],
-      [CONFIG_FILES_PRIORITY, 'filestorage.js', 'routes.js', -1],
-      [CONFIG_FILES_PRIORITY, 'unknown.js', 'sandbox.js', 1],
-      [CONFIG_FILES_PRIORITY, 'log.js', 'sandbox.js', 1],
-      [CONFIG_FILES_PRIORITY, 'sandbox.js', 'sandbox.js', 0],
-      [CONFIG_FILES_PRIORITY, 'log.js', 'log.js', 0],
-      [CONFIG_FILES_PRIORITY, 'tasks.js', 'application.js', -1],
-      [CONFIG_FILES_PRIORITY, 'tasks.js', 'missing_file', -1],
+      testSorting('files.js', 'sandbox.js', 1),
```
belochub: You should not create or call any functions inside the `metatests.case()` call, as I said in https://github.com/metarhia/common/pull/319#discussion_r362078868, just remove the first element in each test case array, and create the function to test by calling `common.sortComparePriority()` only once before the call to `metatests.case()`.

CHANGELOG.md
Bob-chemist: Something like this?

- `sortComparePriority` function contract changed: now it accepts only one argument - `priority` and returns a function, which accepts 2 arguments - filenames. E.g. `sortComparePriority(priority)(file1, file2)`

belochub: First of all, there is no need to repeat the word "changed", it is already specified in the category header above, so just start with

> Contract of `sortComparePriority()` 

So that when its read after the header it combines into:

> Changed contract of `sortComparePriority()`

Secondly, no need to explicitly state that it is a function or a method, the parentheses after the identifier name is a common style that is used to refer to functions/methods in the documentation.

And finally, in your description, you are referring to "filenames", that were never mentioned before neither in the code nor in the documentation of this function. I don't think that the usage of the resulting function is restricted to file names only, so there is probably no need to specify that constraint.

Summarizing the points above, I'd go for something like the following:
- Contract of `sortComparePriority()` to return a function capturing the first argument and accepting the other two.

I'd prefer to get a second opinion on the specific wording of the change though, maybe @lundibundi can chime in.

CHANGELOG.md
```diff
@@ -22,8 +22,7 @@ and this project adheres to
 - `cryptoPrefetcher()` to throw when `bufSize` is not a multiple of
   `valueSize`.
 - `MemoryWritable` internal buffer size is now limited to 8 MB by default.
-- Signature of `callerFilepath()` to allow passing `RegExp` as depth to be used
```
belochub: Why did you delete the existing changelog entry? Revert this change, please.

Bob-chemist: oh, sorry. This happened during rebase. Didn't noticed it

test/sort.js
Bob-chemist: I've found 2 different ways to solve this, but both are not good as for me:

1. By dividing existing `metatests.case()` into 2 parts - with new namespace e.g. `compareFunction`, and another with old `common`.
```
const compareFunction = common.sortComparePriority(CONFIG_FILES_PRIORITY);

metatests.case(
  'Common / sortComparePriority',
  { compareFunction },
  {
    compareFunction: [
      ['files.js', 'sandbox.js', 1],      
      // ...
    ],
  }
);
metatests.case(
  'Common / sort',
  { common },
  {
    'common.sortCompareDirectories': [
      [{ name: '/abc' }, { name: 'abc.ext' }, -1],
      // ...
    ],
  }
);
```

2. By adding a new method to `common` object:
```
common.compareFunction = common.sortComparePriority(CONFIG_FILES_PRIORITY);

metatests.case(
  'Common / sort',
  { common },
  {
    'common.compareFunction': [
      ['files.js', 'sandbox.js', 1],
      ['filestorage.js', 'routes.js', -1],
      // ...
    ],
  }
);
```

belochub: Just pass the function into the object that is the second argument to the `metatests.case()`:
```js
const compareFunction = common.sortComparePriority(CONFIG_FILES_PRIORITY);

metatests.case(
  'Common / sort',
  { common, sortComparePriority: compareFunction },
  {
    sortComparePriority: [
      ['files.js', 'sandbox.js', 1],
      // ...
    ],
    'common.sortCompareDirectories': [
      [{ name: '/abc' }, { name: 'abc.ext' }, -1],
      // ...
    ],
    // ...
  }
);
```

test/sort.js
```diff
@@ -21,19 +21,21 @@ const CONFIG_FILES_PRIORITY = [
...
-      [CONFIG_FILES_PRIORITY, 'tasks.js', 'missing_file', -1],
+    sortComparePriority: [
+      ['files.js', 'sandbox.js', 1],
+      ['filestorage.js', 'routes.js', -1],
+      ['unknown.js', 'sandbox.js', 1],
+      ['log.js', 'sandbox.js', 1],
+      ['sandbox.js', 'sandbox.js', 0],
+      ['log.js', 'log.js', 0],
+      ['tasks.js', 'application.js', -1],
+      ['files.js', 'sandbox.js', 1],
```
belochub: I believe this is a copy-paste mistake, this is probably supposed to be 
```suggestion
      ['tasks.js', 'missing_file', -1],
```

lib/sort.js
```diff
@@ -6,8 +6,8 @@
 //   s2 - <string>, to compare
 // Returns: <number>
```
belochub: Please update the documentation, this function does not return a number.

test/sort.js
```diff
@@ -21,19 +21,21 @@ const CONFIG_FILES_PRIORITY = [
   'routes.js',
 ];
 
+const compareFunction = common.sortComparePriority(CONFIG_FILES_PRIORITY);
```
lundibundi: Nit: I'd just inline the `CONFIG_FILES_PRIORITY` at this point.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to
 - `MemoryWritable` internal buffer size is now limited to 8 MB by default.
 - Signature of `callerFilepath()` to allow passing `RegExp` as depth to be used
   for filtering of stack frames.
+- Contract of sortComparePriority() to return a function capturing the first argument and accepting the other two.
```
lundibundi: Nit: maybe to be a little clearer
```suggestion
- Contract of `sortComparePriority()` to only accept a priority array and return
  a comparison function capturing the first argument and accepting the other two.
```

belochub: @lundibundi I don't think that you still need to mention the first argument later if you expand on its meaning in the first part:

> - Contract of `sortComparePriority()` to only accept a priority array, which was the first argument before the change, and return a comparison function accepting the other two arguments.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to
 - `MemoryWritable` internal buffer size is now limited to 8 MB by default.
 - Signature of `callerFilepath()` to allow passing `RegExp` as depth to be used
   for filtering of stack frames.
+- Contract of `sortComparePriority()` to only accept a priority array, which was the first argument before the change, and return a comparison function accepting the other two arguments.
```
belochub: Wrap the line to 80 characters, please.

lib/sort.js
```diff
@@ -2,12 +2,12 @@
 
 // Compare for array.sort with priority
 //   priority - <string[]>, with priority
+// Returns: <Function>, function(s1, s2)
```
belochub: There is no point to duplicate the signature in place of description:
```suggestion
// Returns: <Function>, comparison function that can be passed to `Array#sort()`
```

lib/sort.js
```diff
@@ -2,12 +2,12 @@
...
 //   s1 - <string>, to compare
 //   s2 - <string>, to compare
```
belochub: You should keep the return type for the returned function:
```suggestion
//   s2 - <string>, to compare
//   Returns: <number>
```

lib/sort.js
```diff
@@ -2,12 +2,15 @@
 
 // Compare for array.sort with priority
 //   priority - <string[]>, with priority
+// Returns: <Function>, comparison function that can be passed to
+// `Array#sort()`
```
belochub: There must be 4 space indentation from the previous line for multi-line descriptions:
```suggestion
//     `Array#sort()`
```

lib/sort.js
```diff
@@ -2,12 +2,15 @@
...
+//
```
belochub: Redundant empty comment line.

PR #323, opened 2020-01-03
CSV stringify module - basic structure

yehorbk:
Basic files (tests and module) created and configured.
Module connected to library.

### DIFF
--- a/lib/csv.js
+++ b/lib/csv.js
@@ -1,0 +1,3 @@
+'use strict';
+
+module.exports = {};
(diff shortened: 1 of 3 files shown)


lib/csv.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const stringify = (data, callback) => {};
```
tshemsedinov:
```suggestion
const stringify = (data, callback) => {
  callback(data);
};
```

tshemsedinov: This small change will prevent tests to fail.

tshemsedinov:
Landed to `csv` branch in https://github.com/metarhia/common/commit/edd2ad1f1d9cc622115ee2294043a3def457a21c

PR #325, opened 2020-03-01
Add ESModules import support

lundibundi:

Also, when 'Conditional exports' are no longer experimental the `package.json` could be changed to use
```json
  "exports": {
    "require": "./common.js",
    "import": "./common.mjs"
  },
```
to automativally resolve '@metarhia/common' depending on `require` or `import` usage. (without the need to use `@metahia/common/module`)

/cc @nechaido @tshemsedinov


tshemsedinov:
I don't like esmodules but let it be :smile:

tools/esmodules-export-gen.js
```diff
@@ -0,0 +1,39 @@
...
+
+const path = require('path');
+const fs = require('fs');
+const common = require('../common');
+
+const COMMON_JS_FILENAME = 'common.js';
+const COMMON_MJS_FILEPATH = './common.mjs';
+
+const header = indexPath =>
+  `// This is an automaticaly generated file. DO NOT MODIFY MANUALLY.
```
SemenchenkoVitaliy:
```suggestion
  `// This is an automatically generated file. DO NOT MODIFY MANUALLY.
```

common.mjs
```diff
@@ -0,0 +1,135 @@
+// This is an automaticaly generated file. DO NOT MODIFY MANUALLY.
```
lundibundi:
```suggestion
// This is an automatically generated file. DO NOT MODIFY MANUALLY.
```

lundibundi:
ping @belochub.

lundibundi:
ping @belochub

tools/esmodules-export-gen.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
```
belochub, line 1: Just a proposal/idea, maybe add a shebang for this file? It's up to you, so if you feel there is no need for it, just skip this suggestion.
```suggestion
#!/usr/bin/env node
'use strict';
```

lundibundi: Done.

belochub:
Landed in https://github.com/metarhia/common/commit/94df32964743067cd7208494b8edb51e7d5cf893.

belochub:
As it turns out, this PR has introduced a regression by breaking support for importing submodules (e.g. `require('@metarhia/common/lib/id')`) in latest Node.js versions. As such, unless another commit fixing this will be introduced before the next minor release, this feature will only be made available in the next major release.

PR #330, opened 2020-06-03
Package maintenance

tshemsedinov:


package.json
```diff
@@ -40,16 +40,16 @@
...
+    "@metarhia/doc": "^0.6.1",
     "eslint": "^7.1.0",
-    "eslint-config-metarhia": "^7.0.0",
+    "eslint-config-metarhia": "^7.0.1",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-import": "^2.20.2",
     "eslint-plugin-prettier": "^3.1.3",
     "metatests": "^0.7.1",
-    "prettier": "1.18.x"
+    "prettier": "^1.18.0"
```
lundibundi, line 53: Prettier is known to break (drastically change the formatting of code) in minor releases that's what we put it at a patch-only level.

lundibundi: It looks like that's exactly what happened due to the updated prettier version, hence the broken CI.

tshemsedinov: I'dd add tests to prettier ignore because `tests/` readability is more important and will fix `lib/` @lundibundi

tshemsedinov: Also I didn't update to the last prettier because it fixes `x => x` to `(x) => x` :smile:

lundibundi: I'd prefer to have such [ignore] rules only for declarative tests to still get proper formatting of imperative tests, we could i.e. just add prettier ignore statements for declarative tests or put them in a separate folder.

lundibundi: > Also I didn't update to the last prettier because it fixes x => x to (x) => x :)

This is fixable, they just changed the default.
https://prettier.io/docs/en/options.html#arrow-function-parentheses

.travis.yml
```diff
@@ -1,16 +1,13 @@
 language: node_js
 node_js:
-  - 10
```
lundibundi, line 3: Why do we remove 10 though (it was an LTS an in maintenance now, at the same time 13 is EOL already)?

LICENSE
```diff
@@ -1,6 +1,7 @@
 MIT License
 
-Copyright (c) 2017-2019 Metarhia contributors (full list in AUTHORS file)
+Copyright (c) 2017-2020 Metarhia contributors
+(https://github.com/metarhia/common/blob/master/AUTHORS).
```
belochub, line 4: `AUTHORS` file is published with the package, so I don't see the point in adding the link to it here.
Besides, it's possible for the repository to move somewhere, thus invalidating this link.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metarhia/common",
-  "version": "2.1.0",
+  "version": "2.1.1",
```
belochub, line 3: You can't make it a patch version since there is new functionality added since the last release.
Besides that, `CHANGELOG.md` wasn't updated to reflect the changes, and overall, I think it would be better to follow a standardized release procedure for this repository by making a separate PR from the `publish` branch with a [`release`](https://github.com/metarhia/common/pulls?q=is%3Apr+label%3Arelease+is%3Aclosed) tag and description of the changes in the PR description extracted from the changelog, instead of silently putting out new version inside a PR containing multiple small changes.

.eslintignore
```diff
@@ -1,3 +1,5 @@
 node_modules/
 coverage/
-dist
+dist/
+test/array.js
+test/data.js
```
belochub, line 5: Why are these files ignored by eslint?

package.json
belochub, line 53: I'd prefer to update Prettier in a separate PR that would also introduce the resulting changes to the code after releasing a new major version of this package without Node.js 8 support.

.travis.yml
```diff
@@ -1,16 +1,13 @@
...
   - 12
   - 13
   - 14
 os:
   - linux
   - osx
   - windows
-
```
belochub, line 11: Is there a reason for removing the empty line separators here and below?

tshemsedinov:
Closed in favour of PR: https://github.com/metarhia/common/pull/333

PR #335, opened 2020-06-05
Update prettier to 1.19.x

lundibundi:
Version prettier@2 is not available due to the need to support (and run
on CI) Node.js 8 for @metahia/common@2.


test/array.js
```diff
@@ -3,6 +3,7 @@
 const metatests = require('metatests');
 const common = require('..');
 
+// prettier-ignore
```
belochub, line 6: Maybe move it to the entries that require specific formatting for improved readability, like `'common.splitAt'`, `'common.range'`, and `'common.sequence'`?

test/data.js
```diff
@@ -5,6 +5,7 @@
 const metatests = require('metatests');
 const common = require('..');
 
+// prettier-ignore
```
belochub, line 8: I don't think this is required here since most of the entries here are already multiline, and it doesn't drastically change readability for the ones that are not.

lundibundi: I'm fine with both, wdyt @tshemsedinov?

belochub:
Landed in https://github.com/metarhia/common/commit/2b4ddbf637bb2811bce2fd3980b0ea7a4715141e.

PR #337, opened 2020-06-19
Allow to wrap iterator with Iterator

lundibundi:
Previously only objects with Symbol.iterator property were supported.
This makes our Iterator only check for .next function on `base`
and only perform `toIterator` if called with `iter` function.

This is a BREAKING CHANGE.

- [x] tests are added/updated
- [ ] documentation is updated (`npm run doc` to regenerate documentation based on comments)

Will update docs and changelog when we reach consensus on this as there are multiple options here.

### DIFF
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -11,7 +11,9 @@ const toIterator = base => {
 
 class Iterator {
   constructor(base) {
-    this.base = toIterator(base);
+    if (typeof base.next !== 'function')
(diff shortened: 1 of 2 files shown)


lundibundi:
Also, while most (all?) Iterators follow the pattern of returning `this` on `[Symbol.iterator]()` call on them the [spec](https://tc39.es/ecma262/#sec-iteration) doesn't mandate that.

PR #338, opened 2020-06-23
Implement iterator#min, iterator#max, iterator#findCompare

lundibundi:
This will simplify the consumption of Iterator in common use-cases.
Allows of searching value based on comparable condition with findCompare
and adds simple wrappers around findCompare to implement min and max
functions.

- [x] tests are added/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to
 - Table of contents to documentation.
 - Get random element from array: `sample(array)`.
 - ECMAScript Modules support via importing '@metarhia/common/module'.
+- `Iterator#min()`, `Iterator#max()`, and `Iterator#findCompare()` to
+  simplify consumption of iterator in common use-cases
+  (finding minimum, maximum, or using a custom condition appropriately).
 
 ### Changed
 
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -199,6 +199,66 @@ class Iterator {
     return result + suffix;
   }
 
+  // Find value in this iterator by comparing every value with
+  // the found one using `comparator`
+  // Signature: comparator[, accessor[, thisArg]]
+  //   comparator <Function> returns `true` if new value should be accepted
+  //     currValue <any> current value, starts with undefined
+  //     nextValue <any> next value
+  //     Returns: <boolean> `true` if next value should be accepted
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default
+  //     value <any> current iterator value
+  //     Returns: <any> value to compare by
+  //   thisArg <any> value to be used as `this` when calling `accessor` and
+  //       `comparator`
+  // Returns: last iterator value where `comparator` returned `true`,
+  //     <undefined> by default
+  findCompare(comparator, accessor, thisArg) {
+    let res = undefined;
+    let resCompareBy = undefined;
+    for (const value of this) {
+      const compareBy = accessor ? accessor.call(thisArg, value) : value;
+      if (comparator.call(thisArg, resCompareBy, compareBy)) {
+        resCompareBy = compareBy;
+        res = value;
+      }
+    }
+    return res;
+  }
+
+  // Find the maximum value in this iterator
+  // Signature: [accessor[, thisArg]]
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default
+  //     value <any> current iterator value
+  //     Returns: <any> value to compare by
+  //   thisArg <any> value to be used as `this` when calling `accessor`
+  // Returns: element with maximum value or <undefined> if iterator is empty
+  max(accessor, thisArg) {
+    return this.findCompare(
+      (curr, next) => curr === undefined || next > curr,
+      accessor,
+      thisArg
+    );
+  }
+
+  // Find the minimum value in this iterator
+  // Signature: [accessor[, thisArg]]
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default
+  //     value <any> current iterator value
+  //     Returns: <any> value to compare by
+  //   thisArg <any> value to be used as `this` when calling `accessor`
+  // Returns: element with minimum value or <undefined> if iterator is empty
+  min(accessor, thisArg) {
+    return this.findCompare(
+      (curr, next) => curr === undefined || next < curr,
+      accessor,
+      thisArg
+    );
+  }
+
   // Call a function with `this`. Will be equivalent to calling `fn(it)`.
   //   fn <Function>
   //     this <Iterator>
--- a/test/iterator.js
+++ b/test/iterator.js
@@ -662,6 +662,111 @@ metatests.testSync('Iterator.chainApply on string', test => {
   test.strictSame(actual, 'h e l l o');
 });
 
+metatests.testSync('Iterator.max no values', test => {
+  const actual = iter([]).max();
+  test.strictSame(actual, undefined);
+});
+
+metatests.testSync('Iterator.max simple', test => {
+  const actual = iter([1, 2, 3, 4, 5]).max();
+  test.strictSame(actual, 5);
+});
+
+metatests.testSync('Iterator.max accessor', test => {
+  const arr = [{ val: 5 }, { val: 4 }, { val: 3 }, { val: 9 }, { val: 4 }];
+  const actual = iter(arr).max(v => v.val);
+  test.strictSame(actual, arr[3]);
+});
+
+metatests.testSync('Iterator.max accessor this', test => {
+  class Arr extends Array {
+    constructor(offset, val) {
+      super(...val);
+      this.offset = offset;
+    }
+
+    getVal(val) {
+      return this.offset * val;
+    }
+  }
+  const arr = new Arr(-1, [5, 4, 3, 2, 1]);
+  const actual = iter(arr).max(arr.getVal, arr);
+  test.strictSame(actual, 1);
+});
+
+metatests.testSync('Iterator.min no values', test => {
+  const actual = iter([]).min();
+  test.strictSame(actual, undefined);
+});
+
+metatests.testSync('Iterator.min simple', test => {
+  const actual = iter([1, 2, 3, 4, 5]).min();
+  test.strictSame(actual, 1);
+});
+
+metatests.testSync('Iterator.min accessor', test => {
+  const arr = [{ val: 5 }, { val: 4 }, { val: 3 }, { val: 9 }, { val: 4 }];
+  const actual = iter(arr).min(v => v.val);
+  test.strictSame(actual, arr[2]);
+});
+
+metatests.testSync('Iterator.min accessor this', test => {
+  class Arr extends Array {
+    constructor(offset, val) {
+      super(...val);
+      this.offset = offset;
+    }
+
+    getVal(val) {
+      return this.offset * val;
+    }
+  }
+  const arr = new Arr(-1, [1, 2, 3, 4, 5]);
+  const actual = iter(arr).min(arr.getVal, arr);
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ and this project adheres to
 - Table of contents to documentation.
 - Get random element from array: `sample(array)`.
 - ECMAScript Modules support via importing '@metarhia/common/module'.
+- `Iterator#min()`, `Iterator#max()` and `Iterator#findCompare()` to
+  allow simpler iterator citerator#minonsumption in common use-cases.
```
belochub, line 22: I don't understand the meaning of "to allow simpler iterator citerator#minonsumtion", can you elaborate?

lib/iterator.js
```diff
@@ -199,6 +199,67 @@ class Iterator {
...
+  // Find value in this iterator compared by `comparator` on values
+  // obtained by applying `accessor` to each value.
+  // Signature: comparator, [accessor[, thisArg]]
+  //   comparator <Function> returns true if new value should be accepted.
+  //     currValue <any> current value, starts with undefined
+  //     nextValue <any> next value
+  //     Returns: <boolean> true if next value should be accepted.
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default.
+  //     value <any> this
```
belochub, line 211: The meaning of 'this' is unclear in this context, you should rewrite this argument's explanation.

lib/iterator.js
```diff
@@ -199,6 +199,67 @@ class Iterator {
...
+  //     Returns: <any> value to compare by.
+  //   thisArg <any> value to be used as `this` when calling `accessor` and
+  //       `comparator`.
+  // Returns: element with maximum `accessor` value.
```
belochub, line 215: It's unclear what 'maximum' value is for complex types, besides that, you can even use this function to find a minimum value, as you do later in this file. You also need to specify that it will return undefined when the value couldn't be found.

lib/iterator.js
```diff
@@ -199,6 +199,67 @@ class Iterator {
...
+    }
+    return res;
+  }
+
+  // Find maximum value in this iterator compared by values obtained by
+  // applying `accessor` to each value.
+  // Signature: [accessor[, thisArg]]
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default.
+  //     value <any> this
```
belochub, line 234: Ditto.

lib/iterator.js
```diff
@@ -199,6 +199,67 @@ class Iterator {
...
+  //     Returns: <any> value to compare by.
+  //   thisArg <any> value to be used as `this` when calling `accessor`.
+  // Returns: element with maximum `accessor` (iterator value by default) value.
```
belochub, line 237: It's unclear that 'iterator value by default' refers to the fact that if a user doesn't pass `accessor` function, values itself will be used for comparison. Apart from that, it's not actually `accessor` value that is being compared since it's the same for every element and is equal to the function that was passed into this method, so you should rewrite this to make it more clear. You also need to specify that it can return undefined for an empty iterator.

lib/iterator.js
```diff
@@ -199,6 +199,67 @@ class Iterator {
...
+      thisArg
+    );
+  }
+
+  // Find minimum value in this iterator compared by values obtained by
+  // applying `accessor` to each value.
+  // Signature: [accessor[, thisArg]]
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default.
+  //     value <any> this
```
belochub, line 251: Ditto

lib/iterator.js
```diff
@@ -199,6 +199,67 @@ class Iterator {
...
+  //     Returns: <any> value to compare by.
+  //   thisArg <any> value to be used as `this` when calling `accessor`.
+  // Returns: element with minimum `accessor` (iterator value by default) value.
```
belochub, line 254: Ditto

test/iterator.js
```diff
@@ -643,6 +643,99 @@ metatests.testSync('Iterator.chainApply on string', test => {
...
+  const arr = [5, 4, 3, 2, 1];
+  const actual = iter(arr).max(
+    function(v) {
+      const val = this.curr;
+      this.curr = v;
+      return val;
+    },
+    { curr: 0 }
+  );
+  test.strictSame(actual, 4);
```
belochub, line 672: Why is the result 4 and not 5 here?

test/iterator.js
```diff
@@ -643,6 +643,99 @@ metatests.testSync('Iterator.chainApply on string', test => {
...
+  const arr = [1, 2, 3, 4, 5];
+  const actual = iter(arr).min(
+    function(v) {
+      const val = this.curr;
+      this.curr = v;
+      return val;
+    },
+    { curr: 10 }
+  );
+  test.strictSame(actual, 2);
```
belochub, line 701: Same question here, I don't understand, how did you get 2 instead of 1?

test/iterator.js
lundibundi: The order is like this
1. value: 5, curr: 0, accessor returned 0
2. value: 4, curr: 5, accessor returned 5 **our max value**
3. value: 3, curr: 4, accessor returned 4
4. value: 2, curr: 3, accessor returned 3
5. value: 1, curr: 2, accessor returned 2

test/iterator.js
lundibundi: The order is like this
1. value: 1, curr: 10, accessor returned 10
2. value: 2, curr: 1, accessor returned 1 **our min value**
3. value: 3, curr: 2, accessor returned 2
4. value: 4, curr: 3, accessor returned 3
5. value: 5, curr: 4, accessor returned 4

lib/iterator.js
```diff
@@ -199,6 +199,68 @@ class Iterator {
...
+  //     nextValue <any> next value
+  //     Returns: <boolean> true if next value should be accepted.
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default.
+  //     value <any> current iterator value.
+  //     Returns: <any> value to compare by.
+  //   thisArg <any> value to be used as `this` when calling `accessor` and
+  //       `comparator`.
+  // Returns: last iterator value where `comparator` returned `true` or
+  //     `undefined` if iterator is empty.
```
belochub, line 216: It's not necessarily `undefined` only in cases when the iterator is empty, it will also be `undefined` if `comparator` didn't return `true` for any of the values.

lib/iterator.js
```diff
@@ -199,6 +199,68 @@ class Iterator {
     return result + suffix;
   }
 
+  // Find value in this iterator compared by `comparator` on values
+  // obtained by applying `accessor` to each value.
```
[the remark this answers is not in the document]
lundibundi: I'd prefer to have them since they are supposed to be a complete sentence, besides those few functions with comments in Iterator mostly have dots at the end.

lib/iterator.js
```diff
@@ -199,6 +199,68 @@ class Iterator {
     return result + suffix;
   }
 
+  // Find value in this iterator compared by `comparator` on values
```
belochub, line 202: Maybe clarify the way it works: 'Find value in this iterator by comparing every value with the found one using `comparator`'

lib/iterator.js
```diff
@@ -199,6 +199,68 @@ class Iterator {
...
+      if (comparator.call(thisArg, resCompareBy, compareBy)) {
+        resCompareBy = compareBy;
+        res = value;
+      }
+    }
+    return res;
+  }
+
+  // Find maximum value in this iterator compared by values obtained by
+  // applying `accessor` to each value
```
belochub, line 231: Just to simplify the description:
```suggestion
  // Find the maximum value in this iterator
```

lib/iterator.js
```diff
@@ -199,6 +199,68 @@ class Iterator {
...
+  max(accessor, thisArg) {
+    return this.findCompare(
+      (curr, next) => curr === undefined || next > curr,
+      accessor,
+      thisArg
+    );
+  }
+
+  // Find minimum value in this iterator compared by values obtained by
+  // applying `accessor` to each value
```
belochub, line 248: Same here:
```suggestion
  // Find the minimum value in this iterator
```

lib/iterator.js
```diff
@@ -199,6 +199,68 @@ class Iterator {
     return result + suffix;
   }
 
+  // Find value in this iterator compared by `comparator` on values
+  // obtained by applying `accessor` to each value
```
lundibundi, line 203:
```suggestion
  // Find value in this iterator by comparing every value with
  // the found one using `comparator`
```

lib/iterator.js
```diff
@@ -199,6 +199,66 @@ class Iterator {
     return result + suffix;
   }
 
+  // Find value in this iterator by comparing every value with
+  // the found one using `comparator`
+  // Signature: comparator[, accessor[, thisArg]]
+  //   comparator <Function> returns true if new value should be accepted
```
SemenchenkoVitaliy, line 205:
```suggestion
  //   comparator <Function> returns `true` if new value should be accepted
```
For consistency with Returns.

lib/iterator.js
```diff
@@ -199,6 +199,66 @@ class Iterator {
...
+  //     currValue <any> current value, starts with undefined
+  //     nextValue <any> next value
+  //     Returns: <boolean> true if next value should be accepted
```
SemenchenkoVitaliy, line 208:
```suggestion
  //     Returns: <boolean> `true` if next value should be accepted
```
Ditto. And I'm also not a fan of description duplication.

lib/iterator.js
```diff
@@ -199,6 +199,66 @@ class Iterator {
...
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default
+  //     value <any> current iterator value
+  //     Returns: <any> value to compare by
+  //   thisArg <any> value to be used as `this` when calling `accessor` and
+  //       `comparator`
+  // Returns: last iterator value where `comparator` returned `true`,
+  //     `undefined` by default
```
SemenchenkoVitaliy, line 216:
```suggestion
  //     <undefined> by default
```

lib/iterator.js
```diff
@@ -199,6 +199,66 @@ class Iterator {
...
+  }
+
+  // Find the maximum value in this iterator
+  // Signature: [accessor[, thisArg]]
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default
+  //     value <any> current iterator value
+  //     Returns: <any> value to compare by
+  //   thisArg <any> value to be used as `this` when calling `accessor`
+  // Returns: element with maximum value or `undefined` if iterator is empty
```
SemenchenkoVitaliy, line 237:
```suggestion
  // Returns: element with maximum value or <undefined> if iterator is empty
```

lib/iterator.js
```diff
@@ -199,6 +199,66 @@ class Iterator {
...
+  }
+
+  // Find the minimum value in this iterator
+  // Signature: [accessor[, thisArg]]
+  //   accessor <Function> gets value to compare by, current iterator value
+  //       is used by default
+  //     value <any> current iterator value
+  //     Returns: <any> value to compare by
+  //   thisArg <any> value to be used as `this` when calling `accessor`
+  // Returns: element with minimum value or `undefined` if iterator is empty
```
SemenchenkoVitaliy, line 253:
```suggestion
  // Returns: element with minimum value or <undefined> if iterator is empty
```

lib/iterator.js
lundibundi: I assume these are to get them to be made into links but I think it would be better to just handle `` `type` `` the same way as ` <type> `

belochub:
Landed in https://github.com/metarhia/common/commit/e83fc7ebd1857a04374e06575895068c8cd34d16.

PR #339, opened 2020-07-02
Implement Iterator#partition() to partition it into arrays

lundibundi:
- [x] code is properly formatted (`npm run fmt`)
- [x] tests are added/updated
- [x] documentation is updated (`npm run doc` to regenerate documentation based on comments)
- [x] description of changes is added under the `Unreleased` header in CHANGELOG.md

PR #340, opened 2020-07-02
Implement Iterator#keepLast() to allow storing last iterator value

lundibundi:
- [x] tests are added/updated

### DIFF
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -222,6 +222,17 @@ class Iterator {
     return iter(res && res[Symbol.iterator] ? res : [res]);
   }
 
+  // Creates an iterator that stores last returned value and allows reinserting
+  // it into the iterator.
+  // Example:
+  // const keeper = iter([1, 2, 3, 4]).keepLast();
+  // keeper.takeWhile(v => v < 3).toArray(); // [1, 2]
+  // keeper.repeatLast();
+  // keeper.toArray(); // [3, 4]
+  keepLast() {
+    return new KeepLastIterator(this);
+  }
+
   // Create iterator iterating over the range
   // Signature: start, stop[, step]
   //   start <number>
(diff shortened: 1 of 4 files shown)


lib/iterator.js
```diff
@@ -490,6 +501,34 @@ class SkipWhileIterator extends Iterator {
...
 }
 
+class KeepLastIterator extends Iterator {
+  constructor(base) {
+    super(base);
+    this.repeat = false;
+    this.lastElement = undefined;
+  }
+
+  get last() {
```
lundibundi, line 511:
```suggestion
  last() {
```

lib/iterator.js
```diff
@@ -490,6 +502,39 @@ class SkipWhileIterator extends Iterator {
   }
 }
 
+class KeepLastIterator extends Iterator {
+  constructor(base) {
+    super(base);
+    this.repeat = false;
+    this.lastElement = undefined;
```
nechaido, line 509: I don't really like `lastElement` sound like instead of `{ value: T, done: boolean }` it is `T`.

lundibundi:
I think the proposed `splitOn` is something like https://github.com/metarhia/common/pull/339.

Will think about a better solution.

> Leading to a situation where you have multiple iterators sharing a common state.

Though, what's wrong with this? I think it's already possible if someone wants to take values from the same iterator in multiple places (which is fine IMO) or misuses iterator (like using `async` function in `.map` etc).

nechaido:
> Though, what's wrong with this? I think it's already possible if someone wants to take values from the same iterator in multiple places (which is fine IMO) or misuses iterator (like using async function in .map etc).

The problem is that it forces programmers to save iterator pre-transformation.
I know that it is currently possible to misuse iterators, but I believe that we should not encourage such patterns.

PR #341, opened 2020-07-03
Implement Iterator#groupBy allowing to group values by keys

lundibundi:
- [x] tests are added/updated

### DIFF
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -214,6 +214,25 @@ class Iterator {
     return new EnumerateIterator(this);
   }
 
+  // Consumes an iterator grouping values by keys
+  // Signature: classifier[, thisArg]
+  //   classifier <Function> gets value to group by
+  //     value <any> current iterator value
+  //     Returns: <any> value to group by
+  //   thisArg <any> value to be used as `this` when calling `classifier`
+  //   Returns: <Map> map with arrays of iterator values grouped
+  //       by keys returned by `classifier`
+  groupBy(classifier, thisArg) {
+    const map = new Map();
+    for (const value of this) {
+      const key = classifier.call(thisArg, value);
+      const collection = map.get(key);
+      if (!collection) map.set(key, [value]);
+      else collection.push(value);
+    }
+    return map;
+  }
+
   join(sep = ',', prefix = '', suffix = '') {
     let result = prefix;
     const { done, value } = this.next();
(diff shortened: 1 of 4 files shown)


lib/iterator.js
```diff
@@ -187,6 +187,23 @@ class Iterator {
     return new EnumerateIterator(this);
   }
 
+  // Consumes an iterator grouping values by keys
+  //   accessor <Function> gets value to group by
+  //     value <any> current iterator value
+  //     Returns: <any> value to group by
+  //   Returns: <Map> map with arrays of iterator values grouped
+  //       by keys returned by `accessor`
+  groupBy(accessor) {
```
nechaido, line 196:
```suggestion
  groupBy(accessor, thisArg) {
```

lib/iterator.js
```diff
@@ -187,6 +187,23 @@ class Iterator {
...
+    const map = new Map();
+    for (const value of this) {
+      const key = accessor(value);
```
nechaido, line 199: Should we?
```js
      const key = accessor?.(value) ?? value;
```

lundibundi: That will make this code incompatible with Node.js <14 which is not acceptable since we currently support >=8

lib/iterator.js
lundibundi: Done

lib/iterator.js
nechaido: Yeah, I know, that is why it is not a ```suggestion.
The question is "Should we make it behave this way?".

lundibundi: Oh, I thought about it initially but got to the conclusion that it must be explicit to be clear what is the grouping factor. Also, I don't think it is very useful to group by iterator value anyway.

nechaido: The main reason is to be consistent with the accessor in `findCompare`, `max` and `min`.

lundibundi: Well, this is a different use-case which IMO doesn't benefit from such default case compared to those other functions.

test/iterator.js
```diff
@@ -767,6 +767,75 @@ metatests.testSync('Iterator.findCompare accessor this', test => {
...
+    )
+  );
+});
+
+metatests.testSync('Iterator.groupBy thisArg', test => {
+  const actual = iter([1, 2, 3, 4, 5]).groupBy(
+    function(v) {
+      return v % this.dis;
+    },
+    { dis: 2 }
```
belochub, line 825: What does `dis` mean here? Can you maybe rename it to have a meaningful name, like `radix` or  `base` for example?

PR #344, opened 2020-07-13
Add TypeScript infrastructure and typings

lundibundi:
- [x] tests are added/updated
- [ ] documentation is updated (`npm run doc` to regenerate documentation based on comments)

Since I always have the Iterator typings in my TS projects anyway I guess we could add them here, I'd support these to the best of my ability.
Not sure what would be the best way to go for other modules but since they rarely change I guess we could just type them as well. I would *not* change the process of PRing changes to include TS as people may not be familiar with it but rather someone updates them as we go forward (preferably before release :smile: ).

/cc @belochub

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "dist/lib/network.js": false,
     "dist/lib/stream.js": false
   },
+  "types": "./types/metarhia__common.d.ts",
   "readmeFilename": "README.md",
   "files": [
     "common.mjs",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -32,16 +32,17 @@
     "dist/lib/network.js": false,
     "dist/lib/stream.js": false
   },
+  "types": "./types/common.d.ts",
   "readmeFilename": "README.md",
   "files": [
     "common.mjs",
     "lib/",
     "dist/"
```
belochub, line 40:
```suggestion
    "dist/",
    "types/"
```

lundibundi:
@tshemsedinov @nechaido @belochub @SemenchenkoVitaliy this is waiting for feedback (in general) to decide whether to go on with it. If we approve this idea/approach I could go on adding typings for other modules in common.
(I guess if you're fine with it you can just :+1: this comment, and otherwise comment)

tshemsedinov:
I believe we need ts typings ASAP, tkanks for this initiative :+1: @lundibundi

PR #347, opened 2020-07-20
Implement Iterator#last to return last value of this iterator

lundibundi:
- [x] tests are added/updated

### DIFF
--- a/lib/iterator.js
+++ b/lib/iterator.js
@@ -234,9 +234,11 @@ class Iterator {
   }
 
   // Consumes an iterator returning last iterator value
+  // Signature: [defaultValue]
+  //   defaultValue <any> value to be used if iterator is empty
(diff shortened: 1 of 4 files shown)


lundibundi:
ping @nechaido @belochub

lib/iterator.js
```diff
@@ -233,6 +233,16 @@ class Iterator {
     return map;
   }
 
+  // Consumes an iterator returning last iterator value
```
SemenchenkoVitaliy, line 236: Maybe
```suggestion
  // Consume an iterator and return last iterator value
```
since most comments use infinitive form?

lundibundi: Current style is consistent with other such functions, we should change this for all of them if we want to.

belochub:
Landed in https://github.com/metarhia/common/commit/827f2513a6ba112924bcfc1f4f84c9cd6fc6156c.

PR #349, opened 2020-07-27
Implement Iterator#zipLongest and improve zip description

lundibundi:
- [x] tests are added/updated

Refs: https://docs.python.org/3.0/library/itertools.html#itertools.zip_longest


lib/iterator.js
```diff
@@ -328,14 +348,30 @@ class Iterator {
...
   }
 
-  // Create iterator by zipping multiple provided iterators into one
-  // Signature: ...iterators
+  // Creates an iterator that aggregates elements from each of the iterators.
+  //
+  // Returns an iterator of Arrays where the i-th tuple contains
+  // the i-th element from each of the passed iterators.
+  // The iterator stops when the shortest input iterable is exhausted.
+  // Signature: base, ...iterators
```
belochub, line 356: It was written this way to specifically hide the implementation detail of having a separate argument which has the same meaning as every other
```suggestion
  // Signature: ...iterators
```

lib/iterator.js
```diff
@@ -328,14 +348,30 @@ class Iterator {
...
   static zip(base, ...iterators) {
     return new ZipIterator(toIterator(base), iterators);
   }
 
+  // Creates an iterator that aggregates elements from each of the iterators.
+  //
+  // If the iterables are of uneven length, missing values are filled-in
+  // with <undefined>. Iteration continues until the longest iterable
+  // is exhausted.
+  // Signature: base, ...iterators
```
belochub, line 368:
```suggestion
  // Signature: ...iterators
```

lib/iterator.js
```diff
@@ -156,10 +156,30 @@ class Iterator {
     return new FlatMapIterator(this, mapper, thisArg);
   }
 
+  // Creates an iterator that aggregates elements from each of the iterators.
+  //
+  // Returns an iterator of Arrays where the i-th tuple contains
+  // the i-th element from each of the passed iterators.
+  // The iterator stops when the shortest input iterable is exhausted.
+  // Signature: ...iterators
```
belochub, line 164: No need for the signature change when the default is correct
```suggestion
```

lib/iterator.js
```diff
@@ -156,10 +156,30 @@ class Iterator {
...
   zip(...iterators) {
     return new ZipIterator(this, iterators);
   }
 
+  // Creates an iterator that aggregates elements from each of the iterators.
+  //
+  // If the iterables are of uneven length, missing values are filled-in
+  // with <undefined>. Iteration continues until the longest iterable
+  // is exhausted.
+  // Signature: ...iterators
```
belochub, line 176: Ditto
```suggestion
```

lib/iterator.js
```diff
@@ -156,10 +156,30 @@ class Iterator {
...
+  //   iterators <Array>
+  // Returns: <Iterator>
+  zipLongest(...iterators) {
```
belochub, line 179: I think we should allow specifying the value that should be used to fill with when there are no values left in one of the iterators.

lib/iterator.js
lundibundi: The metadoc will generate `zip(iterators)`

lib/iterator.js
lundibundi: Yeah, I couldn't devise a way to make a proper API.

belochub (approved):
LGTM with one nit

PR #354, opened 2020-11-09
Implement privatize, private fields on prototypes

tshemsedinov:
Function moved from impress and needed for 
- [x] tests are added/updated

### DIFF
--- a/lib/oop.js
+++ b/lib/oop.js
@@ -23,7 +23,29 @@ const mixin = (target, source) => {
   Object.assign(target, mix);
 };
 
+// Convert instance with public fields to instance with private fields
+//   instance - <Object>, source instance
+// Returns: <Object> - destination instance
+//
+// Example: common.privatize({ private: 5, f() { return this.private; } });
+const privatize = instance => {
+  const iface = {};
+  const fields = Object.keys(instance);
+  for (const fieldName of fields) {
+    const field = instance[fieldName];
+    if (fieldName === fieldName.toUpperCase()) {
+      iface[fieldName] = field;
+    } else if (typeof field === 'function') {
+      const bindedMethod = field.bind(instance);
+      iface[fieldName] = bindedMethod;
+      instance[fieldName] = bindedMethod;
+    }
+  }
+  return iface;
+};
+
 module.exports = {
   override,
   mixin,
+  privatize,
 };
--- a/README.md
+++ b/README.md
@@ -188,6 +188,7 @@ $ npm install @metarhia/common
 - [parseHost](#parsehosthost)
 - [override](#overrideobj-fn)
 - [mixin](#mixintarget-source)
+- [privatize](#privatizeinstance)
 - [Pool](#class-pool)
   - [Pool.prototype.constructor](#poolprototypeconstructorfactory--null)
   - [Pool.prototype.get](#poolprototypeget)
@@ -1333,6 +1334,20 @@ Previous function will be accessible by obj.fnName.inherited
 - `source`: [`<Object>`][object] source methods
 
 Mixin for ES6 classes without overriding existing methods
+
+### privatize(instance)
+
+- `instance`: [`<Object>`][object] source instance
+
+_Returns:_ [`<Object>`][object] - destination instance
+
+Convert instance with public fields to instance with private fields
+
+_Example:_
+
+```js
+common.privatize({ private: 5, f() { return this.private; } });
+```
 
 ### class Pool
 
--- a/common.mjs
+++ b/common.mjs
@@ -90,6 +90,7 @@ export const localIPs = common.localIPs;
 export const parseHost = common.parseHost;
 export const override = common.override;
 export const mixin = common.mixin;
+export const privatize = common.privatize;
 export const Pool = common.Pool;
 export const sortComparePriority = common.sortComparePriority;
 export const sortCompareDirectories = common.sortCompareDirectories;
--- a/test/oop.js
+++ b/test/oop.js
@@ -62,3 +62,18 @@ metatests.test('multiple inheritance with mixin', test => {
   test.strictSame(obj.property5, 'from Child.method3');
   test.end();
 });
+
+metatests.test('privatize', test => {
+  const source = {
+    CONSTANT: 'constant value',
+    field: 'field value',
+    method() {
+      return [this.CONSTANT, this.field];
+    },
+  };
+  const destination = common.privatize(source);
+  test.strictSame(destination.CONSTANT, 'constant value');
+  test.strictSame(destination.field, undefined);
+  test.strictSame(destination.method(), ['constant value', 'field value']);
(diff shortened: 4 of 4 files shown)


lib/oop.js
```diff
@@ -23,7 +23,29 @@ const mixin = (target, source) => {
...
+// Returns: <Object> - destination instance
+//
+// Example: common.privatize({ private: 5, f() { return this.private; } });
+const privatize = instance => {
+  const iface = {};
+  const fields = Object.keys(instance);
+  for (const fieldName of fields) {
+    const field = instance[fieldName];
+    if (fieldName === fieldName.toUpperCase()) {
+      iface[fieldName] = field;
```
Semchik, line 37: Maybe getter would fit better?

lib/oop.js
```diff
@@ -23,7 +23,29 @@ const mixin = (target, source) => {
...
+    } else if (typeof field === 'function') {
+      const bindedMethod = field.bind(instance);
+      iface[fieldName] = bindedMethod;
+      instance[fieldName] = bindedMethod;
+    }
+  }
+  return iface;
```
roman4e, line 44: `return Object.freeze(iface);`

lib/oop.js
```diff
@@ -23,7 +23,29 @@ const mixin = (target, source) => {
...
+//   instance - <Object>, source instance
+// Returns: <Object> - destination instance
+//
+// Example: common.privatize({ private: 5, f() { return this.private; } });
+const privatize = instance => {
+  const iface = {};
+  const fields = Object.keys(instance);
+  for (const fieldName of fields) {
+    const field = instance[fieldName];
+    if (fieldName === fieldName.toUpperCase()) {
```
aleksandr-kuharenko, line 36: Integer properties will not be privatized. Is it expectable?
For example:
```
const obj = { 121: "value" }; 
const privatized = privatize(obj); 

console.log(privatized);
// Output: { '121': 'value' } 
```

a-kuharenko, line 36: Integer properties will not be privatized. Is it expectable?
For example:
```
const obj = { 121: "value" }; 
const privatized = privatize(obj); 

console.log(privatized);
// Output: { '121': 'value' } 
```

lib/oop.js
tshemsedinov: We need even deepFreeze here

lib/oop.js
tshemsedinov: Sure

lib/oop.js
tshemsedinov: Do you have two accounts?

lib/oop.js
tshemsedinov: Done with enumerable getter

lib/oop.js
tshemsedinov: But as of now done with freeze, thanks

aleksandr-kuharenko (approved):

lib/oop.js
```diff
@@ -23,7 +23,36 @@ const mixin = (target, source) => {
...
+// Example: common.privatize({ private: 5, f() { return this.private; } });
+const privatize = instance => {
+  const iface = {};
+  const fields = Object.keys(instance);
+  for (const fieldName of fields) {
+    const field = instance[fieldName];
+    if (typeof field === 'function') {
+      const bindedMethod = field.bind(instance);
+      iface[fieldName] = bindedMethod;
+    } else if (fieldName === fieldName.toUpperCase()) {
```
DScheglov, line 41: Nested `if` doesn't cries about "let's keep constants". What if 
I guess the loop body could be refactored to:

```js
const field = instance[fieldName];
const isMethod = typeof field === 'function';
if (!isMethod && !isConstantName(fieldName)) continue; // skipping private fields
iface[fieldName] = isMethod ? field.bind(instance) : field;
```

lib/oop.js
```diff
@@ -23,7 +23,36 @@ const mixin = (target, source) => {
...
+      const first = fieldName.charCodeAt(0);
+      if (first >= ASCII_A) {
+        Object.defineProperty(iface, fieldName, {
```
DScheglov, line 44: Later `Object.freeze(iface);` covers this property definition.

lib/oop.js
```diff
@@ -23,7 +23,36 @@ const mixin = (target, source) => {
...
+// Returns: <Object> - destination instance
+//
+// Example: common.privatize({ private: 5, f() { return this.private; } });
+const privatize = instance => {
+  const iface = {};
+  const fields = Object.keys(instance);
+  for (const fieldName of fields) {
+    const field = instance[fieldName];
+    if (typeof field === 'function') {
+      const bindedMethod = field.bind(instance);
```
DScheglov, line 39: not `binded`, but `bound`:
```shell
node
> f = () => {};
[Function: f]
> f.bind()
[Function: bound f]
```

lib/oop.js
DScheglov, line 41: What if constant be a function? We define two different approaches for such constants.

lib/oop.js
```diff
@@ -23,7 +23,36 @@ const mixin = (target, source) => {
...
+      if (first >= ASCII_A) {
+        Object.defineProperty(iface, fieldName, {
+          enumerable: true,
+          get: () => field,
+        });
+      }
+    }
+  }
+  return Object.freeze(iface);
+};
```
DScheglov, line 52: what about private methods?
```js
const dest = privatize({
  _privateMethod() {},
  publicMethod() { this._privateMethod(); },
});
```
Many developers could expect that the dest contains only `publicMethod`.
So, I suggest to extract logic of private fields recogintion to the dedicated function and pass it to the `privatize` as a parameter.

```js
const ASCII_A = 'A'.charCodeAt(0);
const ASCII_Z = 'Z'.charCodeAt(0);

const isExplicitlyPrivate = name => name.charAt(0) === '_';

const isMethod = value => typeof value === 'function';

const isConstantName = name => {
  if (name !== name.toUpperCase()) return false;
  const firstCharCode = name.charCodeAt(0);
  return firstCharCode >= ASCII_A && firstCharCode <= ASCII_Z;
};

const isPublicField = (name, value) => isMethod(value) || isConstantName(name);

const isPrivateField = (name, value) =>
  isExplicitlyPrivate(name) || !isPublicField(name, value);

const privatize = (instance, isPrivate = isPrivateField) => {
  const iface = {};
  const fields = Object.keys(instance);
  for (const fieldName of fields) {
    const fieldValue = instance[fieldName];

    if (isPrivate(fieldName, fieldValue)) continue; // skipping private fields

    iface[fieldName] = isMethod(fieldValue)
      ? fieldValue.bind(instance)
      : fieldValue;
  }
  return Object.freeze(iface);
};
```

lib/oop.js
tshemsedinov: Fixed

PR #355, opened 2020-11-10
WIP: Implement deep freeze and freeze for namespaces

tshemsedinov:
- [ ] tests are added/updated

### DIFF
--- a/lib/oop.js
+++ b/lib/oop.js
@@ -23,7 +23,43 @@ const mixin = (target, source) => {
   Object.assign(target, mix);
 };
 
+const REF_TYPES = ['object', 'function'];
+
+// Deep freeze object structure
+//   instance - <Object>, target nested objects structure
+// Returns: <Object> - deep freezed instance
+//
+const deepFreeze = instance => {
+  Object.freeze(instance);
+  const fields = Object.getOwnPropertyNames(instance);
+  for (const fieldName of fields) {
+    const value = instance[fieldName];
+    if (value === null) continue;
+    if (!REF_TYPES.includes(typeof value) || Object.isFrozen(value)) continue;
+    deepFreeze(value);
+  }
+  return instance;
+};
+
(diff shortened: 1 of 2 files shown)


lib/oop.js
```diff
@@ -23,7 +23,43 @@ const mixin = (target, source) => {
...
+// Deep freeze object structure
+//   instance - <Object>, target nested objects structure
+// Returns: <Object> - deep freezed instance
+//
+const deepFreeze = instance => {
+  Object.freeze(instance);
+  const fields = Object.getOwnPropertyNames(instance);
+  for (const fieldName of fields) {
+    const value = instance[fieldName];
+    if (value === null) continue;
```
YegorZaremba, line 37: What if `value === undefined`? I think we should skip it too

YegorZaremba (approved):

lib/oop.js
a-kuharenko: `null` is checked separatelly because `typeof null === 'object'`
`undefined` will be skipped [here](https://github.com/metarhia/common/pull/355/files#diff-3791507a15f8cb71dc1a4f8377ca25674ef714aa9bb9d86995aa186d1ae92925R38)

DScheglov: I guess `value == null` should work well.

lib/oop.js
```diff
@@ -23,7 +23,43 @@ const mixin = (target, source) => {
...
+    if (!REF_TYPES.includes(typeof value) || Object.isFrozen(value)) continue;
+    deepFreeze(value);
+  }
+  return instance;
+};
+
+// Deep freeze namespaces
+//   namespaces - <Object>, target nested objects structure
+//   exceptNames - <Array>, nemes to be skipped for freezing at first level
```
DScheglov, line 46: Parameter name changed from `exceptNames` to `skipNames`

PR #361, opened 2020-12-09
omap method improved

mykolachub:
I improved omap method algorithm and rewrote comments to it.


rodion-arr (changes requested):
Hi @nikolaichub, please run `npm run fmt` in order to proceed with tests running. Currently they failed due to code-style.

